PR #241, opened 2026-04-18
Move sandbox volume attach path to ctld

### LINKED ISSUE #240: Move sandbox volume IO off storage-proxy hot path
## Summary

Sandbox volume IO currently routes every POSIX operation through procd FUSE and storage-proxy gRPC. This makes small-file and metadata-heavy workloads much slower than normal Kubernetes volumes.

This issue tracks the first step of the refactor: keep storage-proxy as the volume control plane, but move sandbox runtime IO toward a node-local mount path so storage-proxy is no longer in the per-read/write/stat hot path.

## Problem

The current runtime path is effectively:

`app -> kernel VFS -> procd FUSE -> gRPC -> storage-proxy -> JuiceFS VFS/cache -> object/meta`

This adds RPC and serialization overhead to operations such as lookup, getattr, open, read, write, flush, fsync, and release. Small files suffer the most because each file requires many metadata and close-path operations.

## Expected behavior

SandboxVolume mounts should preserve dynamic attach semantics for pooled warm sandboxes while removing storage-proxy from the data hot path. The desired direction is:

`app -> kernel VFS -> node-local mounted volume -> backend`

Storage-proxy should remain responsible for volume metadata, auth/session control, snapshot/restore/fork, sync journal, and invalidation coordination.

## Scope

- Add a node-local volume attach abstraction that procd can call during sandbox initialization and runtime mount requests.
- Keep the existing FUSE/gRPC path as a fallback during migration.
- Preserve dynamic mount and unmount for already-running pooled sandboxes.
- Add unit coverage for mount mode selection, fallback behavior, and the node-local attach client contract.

## Non-goals

- Do not remove storage-proxy.
- Do not replace dynamic sandbox mounts with static Kubernetes PVC mounts.
- Do not require recreating warm pods for volume attach.

## Notes

A pure CSI/PVC approach is not sufficient by itself because Kubernetes does not support adding arbitrary new volumeMounts to an already-running warm pod. The refactor should use a CSI-like node-local mount manager or node agent with late bind-mount semantics, while keeping the existing storage-proxy path available until the node-local implementation is fully wired.

laotoutou:
## Summary
- Route procd node-local sandbox volume mounts through ctld instead of a generic node agent, keeping storage-proxy as metadata/API/fallback instead of the runtime file-operation hot path.
- Add ctld volume attach/detach APIs that bind a staged node-local volume into the sandbox mount namespace and detach it by mount session.
- Inject node host IP and ctld port into procd, keep ctld/mount-mode env manager-controlled, and mount the ctld host staging directory in the daemonset.
- Regenerate infra CRDs/deepcopy for the ctld procd config fields.

Closes #240

## Testing
- make manifests
- go test ./ctld/...
- go test ./manager/procd/...
- go test ./manager/cmd/procd
- go test ./manager/pkg/apis/sandbox0/v1alpha1
- go test ./infra-operator/...
- go test ./manager/pkg/service
- go test ./storage-proxy/pkg/volume ./storage-proxy/pkg/grpc ./storage-proxy/pkg/http
- git diff --check
- pre-push hook: make manifests, make proto, make apispec, go fmt ./..., go mod tidy, go mod vendor, golangci-lint run ./...


laotoutou:
Remote kind validation for latest head `14d7b08`:

- Synced the worktree to the kind host and ran `make test-again` from `/root`.
- Build completed for all local binaries, including `ctld`, `procd`, `infra-operator`, and the new `/usr/local/bin/juicefs` helper.
- `sandbox0-system` reached Ready; all fullmode pods were `1/1 Running` with zero restarts:
  - cluster-gateway
  - ctld daemonset pods on both worker nodes
  - manager
  - netd daemonset pods on both worker nodes
  - postgres
  - registry
  - rustfs
  - storage-proxy
- `Sandbox0Infra/fullmode` reported `phase=Ready` and `lastMessage=All services are healthy`.
- Verified `fullmode-ctld` args include:
  - `-volume-staging-root=/host-var/lib/sandbox0/volumes`
  - `-volume-cache-root=/host-var/lib/sandbox0/juicefs-cache`
  - `-volume-config=/config/storage-proxy/config.yaml`
  - `-juicefs-bin=/usr/local/bin/juicefs`
- Verified the ctld image contains `/usr/local/bin/juicefs` (`118M`).
- Checked ctld logs: health/readiness are returning 200 and there is no `JuiceFS volume staging disabled` startup message.

Also fixed one issue found during remote validation: ctld originally reconciled before the storage-proxy ConfigMap existed, so the config mount blocked pod creation. The operator workflow now reconciles storage-proxy before ctld/fuse-device-plugin, and tests cover that ordering plus the ctld config mount.

laotoutou:
Additional remote kind validation with `s0` on latest head `14d7b08` found a functional failure in the node-local path.

What I tested:

1. Patched remote `Sandbox0Infra/fullmode` to `spec.services.manager.config.procdConfig.mountMode=node-local`.
2. Verified the new manager config was live:
   - `mountMode: node-local`
   - `nodeLocalFallbackToStorage: false`
3. Used `s0` against the remote kind cluster:
   - builtin login succeeded
   - template `default` was present
   - sandbox create succeeded
   - volume create succeeded
   - direct volume file write succeeded (`/seed.txt`)
   - dynamic mount failed via `s0 sandbox volume mount ...`

Observed failure:

- CLI error:
  - `sandbox0 API error (500): mount_failed - attach node-local volume through ctld: missing sandbox id for ctld volume attach`
- `cluster-gateway` logs show the failing request reached procd mount API:
  - `POST /api/v1/sandboxvolumes/mount` -> `500`
  - upstream target was the sandbox procd endpoint
- The claimed sandbox pod had procd env consistent with the new path:
  - `mount_mode=node-local`
  - `node_local_fallback_to_storage=false`
- I also observed `ctld_base_url=null` in the procd env dump for that sandbox pod, which looks suspicious and likely needs a separate fix or confirmation.

So the current state is:

- cluster deployment is healthy
- direct volume API works
- node-local dynamic sandbox mount is still broken in live validation

This means PR #241 is not yet functionally complete for the `s0 sandbox volume mount` flow.

laotoutou:
Updated this branch with the node-local-only refactor and a working ctld attach path.

What changed:
- removed procd mount-mode branching; sandbox volume attach now only uses ctld-backed node-local attach
- deleted the old storage-proxy hot-path file I/O/watch implementation from procd (`grpc_fs.go`, `watch.go`, related tests)
- kept storage-proxy for control-plane / volume metadata duties, but moved sandbox mount work to ctld
- fixed `tools/juicefs` mount config so node-local stage mounts no longer panic on nil `FuseOpts`
- ctld now stages JuiceFS node-local mounts and attaches them into the sandbox by:
  - resolving the sandbox process cgroup
  - preferring the procd process over the pause process
  - entering the sandbox mount namespace
  - setting cwd to `/proc/<procd-pid>/root`
  - using `open_tree + move_mount` onto the relative mount target
- removed the public/user-facing procd mount mode config knobs so only node-local remains
- added/updated focused tests around procd identity propagation, node-local attach behavior, ctld process selection, and JuiceFS config construction

Validation:
- local: `go test ./ctld/internal/ctld/volume ./tools/juicefs ./manager/procd/... ./infra-operator/... ./manager/pkg/apis/sandbox0/v1alpha1`
- remote kind (`sandbox0-voltest`): passed real smoke after restoring operator management
  - `volume files write`
  - `sandbox volume mount`
  - sandbox read from mounted file
  - sandbox write through mounted path
  - `volume files cat` sees the sandbox-written file

Remote note: I had to prune stale images inside the kind nodes a few times because repeated image loads filled containerd storage and started failing with `no space left on device`. The final verification was rerun after re-enabling `infra-operator` and succeeded in the normal managed state.

laotoutou:
Final validation update on commit `300b9a8`:

- local targeted tests passed:
  - `go test ./ctld/internal/ctld/volume ./tools/juicefs ./manager/procd/... ./infra-operator/... ./manager/pkg/apis/sandbox0/v1alpha1`
- remote kind smoke passed twice on `sandbox0-voltest`
  - once with the DaemonSet pinned to the unique ctld test image
  - once again after loading `sandbox0ai/infra:latest`, re-enabling `infra-operator`, and letting the normal managed state converge

Final remote smoke flow:
- create sandbox
- create volume
- `volume files write /seed.txt`
- `sandbox volume mount --path /workspace/data`
- sandbox reads mounted file
- sandbox writes through mounted path
- `volume files cat` reads back the sandbox-written content

The key ctld-side fix was to attach into the sandbox mount namespace using the sandbox `procd` pid, set cwd to `/proc/<procd-pid>/root`, and `move_mount` the staged node-local JuiceFS mount onto the relative target path.

laotoutou:
Benchmark update from remote kind (`sandbox0-voltest`, Ubuntu arm64):

Scenario: inside one sandbox, compare local `/tmp` vs mounted SandboxVolume `/workspace/data`; 2000 files x 1 KiB, 3 rounds; each round does create/write/close, read all files, and unlink files. No fsync. Values below are medians.

| path | write median | write ops/s | read median | read ops/s | unlink median | unlink ops/s |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| `/tmp` local overlay | 0.564s | 3544 | 0.060s | 33312 | 0.118s | 16930 |
| `/workspace/data` SandboxVolume | 92.697s | 21.6 | 20.092s | 99.5 | 40.625s | 49.2 |

Slowdown vs `/tmp`: write ~164x, read ~335x, unlink ~344x.

Interpretation: this PR successfully moves file I/O out of `storage-proxy` and verifies dynamic node-local mounts work, but the benchmark shows the remaining hot bottleneck is now the JuiceFS small-file metadata/data path in the remote kind deployment, not `procd -> storage-proxy` gRPC. If we want SandboxVolume small-file performance to approach normal Kubernetes local volume behavior, the next architectural step should likely be a node-local writable staging/cache layer with async/checkpointed persistence to JuiceFS, rather than direct small-file operations against JuiceFS for every file.

PR #243, opened 2026-04-19
Replace procd volume gRPC path with s0vp

laotoutou:
## Summary
- Replace the procd-to-storage-proxy volume hot path with the `s0vp` TCP binary protocol.
- Remove the storage-proxy gRPC server, generated gRPC stubs, gRPC auth interceptor, and procd gRPC session client.
- Wire operator config to `volumeProtocolAddr`/`volumeProtocolPort` and keep HTTP separate from the mount path.
- Add primary redirect support, custom storage-proxy filesystem errors, s0vp codec tests, and the `procd.md` design note.

## Testing
- `make manifests`
- `make proto`
- `go test ./storage-proxy/pkg/fserror ./storage-proxy/pkg/fsserver ./storage-proxy/pkg/http ./storage-proxy/pkg/volumeserver ./storage-proxy/pkg/volproto ./tests/integration/internal/tests/storage-proxy ./tests/integration/internal/tests/manager`
- `go test $(go list ./... | rg -v 'tests/e2e/scenarios/single-cluster')`\n- `golangci-lint run ./...` via pre-push hook\n- Attempted `go test ./...`; local kind e2e cannot start because Docker daemon is unavailable.


laotoutou:
Remote kind mounted-volume benchmark after replacing the procd volume gRPC path with s0vp and fixing mutation invalidation remounts.

Workload: 500 files, 256 bytes per file, executed inside the sandbox pod container.

| Path | Write | Read | list+stat |
| --- | ---: | ---: | ---: |
| pod local /tmp | 5,663.9 ops/s | 9,590.3 ops/s | 26,740.7 ops/s |
| mounted s0fs volume /workspace/bench-volume | 54.7 ops/s | 262.4 ops/s | 6,120.7 ops/s |

Previous baseline for the same workload was 47.5 write ops/s, 233.8 read ops/s, and 8,244 list+stat ops/s on mounted s0fs. Removing gRPC/protobuf from the mounted-volume hot path improves write/read slightly but does not change the order of magnitude. The remaining bottleneck is now the per-file FUSE request pattern plus storage-proxy's synchronous small-file mutation path, not HTTP or protobuf serialization.

Correctness note: the first remote benchmark exposed an EIO caused by treating every pathless S0FS mutation invalidation as a full remount. Commit d4b21ab changes remounts to only happen for explicit global invalidations with invalidate_id, and normal mutation events now use kernel inode cache notify/watch semantics without disrupting an active FUSE mount.

laotoutou:
Remote kind benchmark after `245a5ed` (`Use local WAL durability for RWO s0fs writes`). Workload: 500 files, 256 bytes per file, benchmark executed inside the sandbox pod `procd` container.

| Path | Write | Read | list+stat |
| --- | ---: | ---: | ---: |
| pod local `/tmp` | 5,640.9 ops/s | 5,788.1 ops/s | 50,559.4 ops/s |
| mounted s0fs volume `/workspace/bench-volume` | 91.5 ops/s | 252.9 ops/s | 6,887.3 ops/s |

Compared with the previous remote run after removing gRPC (`54.7 write / 262.4 read / 6,120.7 list+stat`), removing synchronous RWO sync-journal recording improved small-file write by about 67%, but it is still far from the 10x target. The direct in-process S0FS engine remains orders of magnitude faster (`~43,988 write ops/s`, `~537,471 read ops/s` from the earlier direct engine probe), so the remaining bottleneck is the mounted execution path: FUSE fanout plus one `s0vp` round trip per small POSIX operation.

Conclusion: this PR removes gRPC and makes RWO writes acknowledge on the local S0FS WAL, but the next order-of-magnitude design should move the RWO filesystem execution/WAL path to pod/node local and make storage-proxy the coordination and async replication plane. RWX should stay on the centralized correctness path.

laotoutou:
Scope update: the `245a5ed` local-WAL/RWO sync-journal experiment has been removed from this PR via force-with-lease. Current PR head is `d4b21ab` and includes only the gRPC removal/s0vp protocol work plus the remount/EIO fix.

The benchmark I posted for `245a5ed` should be read only as a diagnostic experiment: it showed synchronous journal removal improves write throughput but is not enough for the 10x target. It is not part of the current PR diff.

PR #253, opened 2026-04-20
Optimize s0fs materialization and cold reads

### LINKED ISSUE #252: Optimize s0fs materialization and cold-read path for production volume performance
## Context

The current s0fs materializer is safe enough for basic persistence, but it has several production performance and operability risks:

- Background materialization coalesces writes, but each successful materialize rewrites the current full file-data state into a new segment.
- Old manifests and segments are not reclaimed by the s0fs materializer path.
- Cold reads for many small files issue one object-store range read per file even when the files live in the same segment.
- Snapshot/fork and owner handoff paths need stronger tests around materialize visibility.

## Goals

- Avoid full-segment rewrite for unchanged cold data during materialize.
- Keep current-manifest reads efficient for small-file workloads.
- Add GC or reachable-object discipline for materialized segments/manifests.
- Preserve POSIX correctness and snapshot/fork correctness.
- Add unit, integration, and e2e coverage for mounted volume writes, API visibility, materialize reload, and cold small-file reads.

## Non-goals

- Do not reintroduce JuiceFS.
- Do not route the mounted volume hot path back through storage-proxy.
- Do not weaken RWO correctness for performance.

## Validation

- Unit tests for s0fs materializer/read path.
- Storage-proxy/ctld integration tests where applicable.
- Remote kind e2e covering mounted volume write/read and volume API visibility.

laotoutou:
## Summary

Closes #252.

This PR changes s0fs materialization from full data rewrite on every dirty flush to a retained-cold-data model:

- retain unchanged cold file extents across materializations instead of hydrating and rewriting them;
- write a new segment only for hot/modified file data;
- switch the owner engine to the materialized cold manifest state after a successful materialize;
- add a bounded per-engine segment cache so small-file cold reads from the same packed segment avoid one S3 range GET per file;
- keep snapshot/fork/export paths using inline hydrated state so snapshots and forks do not reference source-volume segments;
- extend e2e mounted-volume coverage with a small-file batch written inside the sandbox and read back through the volume API.

## Tests

- `go test ./storage-proxy/pkg/s0fs`
- `go test ./storage-proxy/pkg/snapshot`
- `go test ./storage-proxy/pkg/fsserver ./storage-proxy/pkg/http`
- `go test ./tests/e2e/cases ./tests/e2e/utils`
- `go test ./ctld/internal/ctld/...`
- `go test ./storage-proxy/...`
- `go test ./...` partially passed all non-e2e packages; local e2e failed before running specs because local Docker is unavailable: `Cannot connect to the Docker daemon at unix:///var/run/docker.sock`.
- pre-push hook passed: manifests/proto/apispec/fmt/tidy/vendor/golangci-lint.

### DIFF
--- a/storage-proxy/pkg/s0fs/engine.go
+++ b/storage-proxy/pkg/s0fs/engine.go
@@ -6,7 +6,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"io"
 	"os"
 	"slices"
 	"sync"
@@ -498,7 +497,7 @@ func (e *Engine) SyncMaterialize(ctx context.Context) (*Manifest, error) {
 		return nil, nil
 	}
 	version := e.mutationVersion
-	state, err := e.exportStateLocked()
+	state, err := e.materializeStateLocked()
 	if err != nil {
 		e.mu.RUnlock()
 		return nil, err
@@ -514,6 +513,9 @@ func (e *Engine) SyncMaterialize(ctx context.Context) (*Manifest, error) {
 	e.mu.Lock()
 	defer e.mu.Unlock()
 	if e.mutationVersion == version {
+		if manifest.State != nil {
+			e.replaceStateLocked(cloneState(manifest.State))
+		}
 		if err := e.persistCurrentStateLocked(); err != nil {
 			return nil, err
 		}
@@ -959,6 +961,19 @@ func (e *Engine) exportStateLocked() (*SnapshotState, error) {
 	return state, nil
 }
 
+func (e *Engine) materializeStateLocked() (*SnapshotState, error) {
+	state := cloneState(e.currentStateLocked())
+	if len(state.ColdFiles) == 0 {
+		return state, nil
+	}
+	for inode := range state.ColdFiles {
+		if state.Nodes[inode] == nil {
+			delete(state.ColdFiles, inode)
+		}
+	}
+	return state, nil
+}
+
 func (e *Engine) readFileLocked(node *Node, inode uint64, offset uint64, size uint64) ([]byte, error) {
 	if node == nil {
 		return nil, ErrNotFound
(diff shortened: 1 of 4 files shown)


laotoutou:
Remote kind validation passed for the volume scenario.

Command run on the remote kind host:

```bash
cd /root/infra && \
  unset http_proxy https_proxy all_proxy && \
  GOWORK=off \
  E2E_USE_EXISTING_CLUSTER=true \
  E2E_SINGLE_CLUSTER_SCENARIOS=volumes \
  go test -v -count=1 ./tests/e2e/scenarios/single-cluster -timeout=30m
```

Result:

```text
Ran 12 of 48 Specs in 185.562 seconds
SUCCESS! -- 12 Passed | 0 Failed | 0 Pending | 36 Skipped
PASS
ok github.com/sandbox0-ai/sandbox0/tests/e2e/scenarios/single-cluster 185.590s
```

This run includes the updated claim-mounted volume e2e path with a small-file batch written inside the sandbox mount and read back through the volume API.

laotoutou:
Remote kind follow-up after merge.

Earlier remote validation was not only a unit check: I rebuilt/redeployed the PR branch into the remote kind cluster with `make test-again`, then ran the targeted volumes e2e scenario:

```text
E2E_USE_EXISTING_CLUSTER=true E2E_SINGLE_CLUSTER_SCENARIOS=volumes go test -v -count=1 ./tests/e2e/scenarios/single-cluster -timeout=30m
12 Passed | 0 Failed | 36 Skipped
```

After merge, I reran the remote kind deployment and added pod-local mounted-volume benchmarks from inside a claimed sandbox pod. The mounted path was verified as `fuse.sandbox0-volume`:

```text
sandbox0-volume-portal on /workspace/bench-volume type fuse.sandbox0-volume
```

Workload: 500 files, 256 bytes/file, 3 rounds median, measured inside the sandbox pod. Each round writes files, runs `sync`, then reads and list+stats. This script counts parent `mkdir -p` in the write phase, which is the more conservative benchmark variant.

| Path | Write ops/s | Read ops/s | list+stat ops/s |
| --- | ---: | ---: | ---: |
| pod local `/tmp` | 9,728.7 | 44,824.4 | 179,566.1 |
| mounted s0fs volume `/workspace/bench-volume` | 1,452.6 | 5,728.7 | 48,445.5 |

For reference, I also ran a variant with directories pre-created before timing writes:

| Path | Write ops/s | Read ops/s | list+stat ops/s |
| --- | ---: | ---: | ---: |
| pod local `/tmp` | 15,471.8 | 31,333.6 | 171,467.2 |
| mounted s0fs volume `/workspace/bench-volume` | 1,692.4 | 5,815.6 | 98,293.3 |

Compared directionally with the previous thread baseline for mounted s0fs (`write 1,269.7`, `read 1,981.7`, `list+stat 13,889.5` ops/s), the conservative rerun shows:

- Write: ~1.1x
- Read: ~2.9x
- list+stat: ~3.5x

The read/list improvements match the intended effect of avoiding repeated cold segment range reads and retaining cold materialized state. Write remains primarily bounded by the mounted FUSE/ctld/s0fs write path rather than this PR's cold-read/materialization changes.

PR #257, opened 2026-04-20
Fix e2e binary response contract validation

laotoutou:
## Summary

This PR fixes the e2e OpenAPI contract validator for raw binary responses:

- keep `application/octet-stream` payloads as strings even when the payload is empty;
- allow empty binary file reads to validate against `type: string, format: binary` instead of being normalized to `null`;
- add e2e utils coverage for empty and non-empty SandboxVolume file reads.

This was hit while running the remote kind volume e2e after the dirty-close WAL sync change: the API returned a valid `application/octet-stream` response, but the test validator normalized the empty body to `null` and failed schema validation before the scenario could continue.

## Tests

- `go test ./tests/e2e/utils`
- `go test ./storage-proxy/pkg/s0fs ./storage-proxy/pkg/fsserver ./ctld/internal/ctld/portal`
- pre-push hook passed: manifests/proto/apispec/fmt/tidy/vendor/golangci-lint

### DIFF
--- a/tests/e2e/utils/contract_helpers.go
+++ b/tests/e2e/utils/contract_helpers.go
@@ -176,14 +176,8 @@ func normalizePayload(contentType string, payload any) (any, error) {
 	if !isJSONContentType(contentType) {
 		switch value := payload.(type) {
 		case []byte:
-			if len(value) == 0 {
-				return nil, nil
-			}
 			return string(value), nil
 		case string:
-			if strings.TrimSpace(value) == "" {
-				return nil, nil
-			}
 			return value, nil
 		default:
 			return payload, nil
--- a/tests/e2e/utils/contract_helpers_test.go
+++ b/tests/e2e/utils/contract_helpers_test.go
@@ -1,0 +1,32 @@
+package utils
+
+import (
+	"net/http"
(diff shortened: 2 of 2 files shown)


laotoutou:
Remote kind validation completed on the updated branch.

Commands run on `kind`:

```bash
cd /root/infra
make docker-build-local
kind load docker-image sandbox0ai/infra:latest --name sandbox0-e2e
unset http_proxy https_proxy all_proxy
GOWORK=off E2E_USE_EXISTING_CLUSTER=true E2E_SINGLE_CLUSTER_SCENARIOS=volumes \
  go test -v -count=1 ./tests/e2e/scenarios/single-cluster -timeout=30m
```

Result:

```text
Ran 12 of 48 Specs in 179.676 seconds
SUCCESS! -- 12 Passed | 0 Failed | 0 Pending | 36 Skipped
ok github.com/sandbox0-ai/sandbox0/tests/e2e/scenarios/single-cluster 179.704s
```

The previous remote failure was the e2e contract validator normalizing an empty `application/octet-stream` response to `null`; this PR fixes that path and the `volumes` scenario now reaches and passes the volume API tests.

Issue #54, opened 2026-03-18
Update skills docs for credential sources and egress rules


laotoutou:
## Summary
Update `skills/sandbox0` documentation to match the credential-source and network egress contract introduced in PR #51.

## Problem
PR #51 changed the public contract from `egress.authRules[].authRef` to a two-part model:
- sandbox-scoped `credentials.bindings`
- `egress.rules[].credentialRef`

It also added manager CRUD endpoints for `/api/v1/credential-sources`.

The `skills/sandbox0` docs still describe the old `authRules/authRef` model, so the examples and field descriptions are now incorrect.

Affected docs include:
- `skills/sandbox0/references/docs-src/sandbox/network/page.mdx`
- `skills/sandbox0/references/docs-src/self-hosted/configuration/page.mdx`

## Expected behavior
The skill docs should describe the current API and examples should use the new field names and concepts.

## Proposed updates
- Replace `authRules` with `egress.rules` in sandbox network docs.
- Document `credentials.bindings` and how `credentialRef` resolves against bindings.
- Add a short explanation of credential sources and the `/api/v1/credential-sources` API.
- Update self-hosted notes so the per-policy gate description matches the new contract.

## Why it matters
The skill is intended to be self-contained and authoritative for agent integrations. Outdated examples here will lead users and agents to generate invalid requests.

PR #262, opened 2026-04-21
s0fs: move committed heads from object storage to postgres

### LINKED ISSUE #261: s0fs v2: replace mutable S3 head with WAL-first engine and PG-backed volume head
## Summary

`s0fs` v1 currently treats object storage as both the capacity layer and the hot metadata/control-plane path. That design is now the main blocker for production-grade SandboxVolume stability and performance.

The concrete failure mode already showed up in managed-agents as GCS `429 rateLimitExceeded` on `s0fs/manifests/latest.json`, but the underlying problem is broader:

- every materialization overwrites the same mutable object (`manifests/latest.json`);
- HTTP volume mutations can force synchronous materialization on every request;
- cold-file writes can require whole-file fetch before modification;
- metadata is modeled as a whole-tree manifest instead of a commit/checkpoint model;
- object storage ends up serving as the hot head store, which creates single-key hotspots and poor scale-out behavior.

Volume is a core Sandbox0 capability. We need to make the storage engine production-safe before wider rollout.

## Goals

Build `s0fs v2` with these production properties:

1. `s0fs` engine remains independent from PostgreSQL.
2. PostgreSQL is used only by the upper volume layer for head/lease/routing/fencing metadata.
3. Object storage stores immutable blobs only; no hot mutable `latest.json` object.
4. WAL-first write path: user-visible write durability is local WAL durability, not object-store round trips.
5. Owner routing and commit/head management are separated from the storage engine.
6. The design must scale for both:
   - large numbers of small files;
   - large files with partial writes and range reads.
7. The final implementation must be covered by strong unit, integration, and e2e tests.

### 1. Mutable S3/GCS head hotspot

Current materialization writes:

- `manifests/<seq>.json`
- `manifests/latest.json`

The second write is a single-object hotspot and already hits GCS mutation limits.

### 2. Small-file write amplification

A sequence of small HTTP file writes can produce repeated materializations and repeated remote metadata commits.

### 3. Large-file partial-write inefficiency

For cold files, a small modification can require reading the whole file back from object storage before applying the change.

### 4. Whole-tree metadata rewrite

The current manifest serializes the whole logical tree (`Nodes`, `Children`, `ColdFiles`, `Segments`) as one state blob.

### 5. Read path depends on remote mutable head

Refresh/read-consistency currently depends on loading `latest.json`, which turns object storage into the hot coordination path.

### 6. Owner routing and durability are too coupled

Mounted owners, temporary/direct HTTP mounts, and materialization timing are tightly coupled in ways that make correctness and performance trade off against each other.

### A. `s0fs` engine layer

Responsibilities:

- node-local WAL
- in-memory inode / directory / extent state
- chunk / pack encoding
- read cache / readahead / compaction support
- build immutable commit/checkpoint artifacts

Non-responsibilities:

- PostgreSQL access
- lease election
- fencing
- routing
- owner registry

### B. volume control-plane layer

Backed by PostgreSQL.

Responsibilities:

- committed head
- lease / epoch / fencing
- owner routing
- mount heartbeat / owner registry
- GC root discovery
- commit CAS / monotonic head advancement

### C. object storage layer

Stores immutable objects only:

- `chunks/...`
- `packs/...`
- `commits/...`
- `checkpoints/...`

No mutable `latest.json` or other hot head object.

### Mounted owner / direct owner

- `write()` success: WAL append + in-memory apply succeeded
- `fsync()` / dirty `close()` success: WAL durable on local disk
- object-store flush: asynchronous batch commit
- committed-head advancement: handled by the volume layer

## Desired read semantics

- owner reads from local hot state/cache
- non-owner opens from committed head
- large files read by chunk/range
- small files prefer pack/cache hits
- no dependence on mutable object-store head keys

### Phase 1: immutable manifest head + PG-backed volume head

This is the first production slice and should land first.

- remove `manifests/latest.json` from the hot path;
- write only immutable manifest objects;
- move committed-head lookup to the volume layer via PostgreSQL;
- teach `s0fs` to load committed head through an injected interface rather than from mutable object storage;
- add strong tests that prove repeated materialization no longer mutates a single shared S3/GCS key.

### Phase 2: WAL-first write path and commit scheduling

- make batch/idle/unmount driven materialization the default commit path;
- keep durability on WAL sync, not object-store sync;
- ensure owner routing remains correct for HTTP and mounted volume users.

### Phase 3: data-layout improvements

- small-file pack objects
- large-file chunk objects
- partial-write / range-read friendly layout

### Phase 4: checkpoint/delta metadata model

- move beyond whole-tree state blob
- checkpoint + delta commit model
- prepare for better compaction and GC

## Acceptance criteria for this workstream

1. No production path depends on overwriting `manifests/latest.json`.
2. Object storage head lookup is removed from the hot path.
3. Mounted volumes and HTTP volume APIs continue to satisfy current correctness expectations.
4. The resulting system is covered by unit, integration, and e2e tests for volume stability.
5. Remote kind validation passes before rollout.

## Notes

This issue tracks the full `s0fs v2` production design. The implementation may land in internal sub-phases, but rollout should converge through one release-quality PR with complete tests and remote kind validation because volume CI/e2e/deploy cycles are expensive and the feature is production-critical.

laotoutou:
Closes #261

## Summary
- add a committed-head abstraction to `s0fs` so the engine no longer depends on mutable object-store head updates in the hot path
- persist committed S0FS manifest heads in PostgreSQL with compare-and-swap semantics and wire storage-proxy/ctld mounts through that head store
- keep immutable manifest objects in object storage, bootstrap legacy `manifests/latest.json` into PG on first read, and add unit/integration coverage for head advancement and conflict detection

## Why
`issue #54` was exposing a deeper S0FS v1 problem: we were treating object storage as the hot committed head store by repeatedly overwriting `manifests/latest.json`. This PR moves the committed pointer into PG while keeping `s0fs` itself PG-agnostic through an injected interface.

## Tests
- `go test ./storage-proxy/pkg/s0fs`
- `go test ./storage-proxy/pkg/db`
- `go test ./storage-proxy/pkg/volume`
- `go test ./storage-proxy/cmd/storage-proxy ./ctld/internal/ctld/portal`
- `go test ./tests/integration/internal/tests/storage-proxy`
- `go test ./tests/e2e/utils`
- remote kind deploy succeeded via `/root/Makefile:test-again`
- remote kind manual volume validation in progress

### DIFF
--- a/ctld/internal/ctld/portal/manager.go
+++ b/ctld/internal/ctld/portal/manager.go
@@ -43,6 +43,7 @@ type Manager struct {
 	mu              sync.Mutex
 	portals         map[string]*portalMount
 	portalsByTarget map[string]*portalMount
+	boundVolumes    map[string]*boundVolume
 	volumes         *localVolumeManager
 }
 
@@ -59,6 +60,15 @@ type portalMount struct {
 	volumeID  string
 	teamID    string
 	mountedAt time.Time
+}
+
+type boundVolume struct {
+	volumeID  string
+	teamID    string
+	access    volume.AccessMode
+	mountedAt time.Time
+	refCount  int
+	volCtx    *volume.VolumeContext
 
 	heartbeatCancel context.CancelFunc
 	heartbeatDone   chan struct{}
@@ -109,6 +119,7 @@ func NewManager(cfg Config) *Manager {
 		heartbeatInterval: heartbeatInterval,
 		portals:           make(map[string]*portalMount),
 		portalsByTarget:   make(map[string]*portalMount),
+		boundVolumes:      make(map[string]*boundVolume),
 		volumes:           newLocalVolumeManager(),
 	}
 	manager.volumeAPI = newMountedVolumeAPIHandler(storageConfig, cfg.Repository, manager.volumes, l)
@@ -234,6 +245,10 @@ func (m *Manager) Bind(ctx context.Context, req ctldapi.BindVolumePortalRequest)
 	if err != nil {
 		return ctldapi.BindVolumePortalResponse{}, err
 	}
+	accessMode, err := validateBindableAccessMode(volumeRecord.AccessMode)
+	if err != nil {
+		return ctldapi.BindVolumePortalResponse{}, err
+	}
 
 	key := portalKey(req.PodUID, portalName)
 	m.mu.Lock()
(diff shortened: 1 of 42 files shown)


laotoutou:
Remote kind update on top of `d399d97` (`ctld: support rox shared volume portals`).

What changed in this follow-up:
- fixed `ctld` volume-portal bind so `ROX` volumes are accepted by the node-local portal path instead of being hard-blocked as `RWO`-only
- added shared bound-volume lifecycle in `ctld` so multiple same-node `ROX` portals reuse one local engine / owner registration / materializer and unbind with ref-counted teardown
- added unit coverage for bindable access-mode validation and shared unbind ref-counting
- added e2e cases for `ROX` shared read-only mounts and `RWX` sandbox-mount rejection

Local verification:
- `go test ./ctld/internal/ctld/portal`
- `go test ./manager/pkg/service`
- `go test ./tests/e2e/cases`

Remote kind manual validation after redeploy:
- `RWO` regression path still passes:
  - unmounted API write/read
  - first sandbox mount succeeds
  - second `RWO` mount is rejected with conflict
  - sandbox write -> volume API read works
  - volume API write -> mounted sandbox read works
  - mounted snapshot create is rejected
  - after sandbox delete + wait-for-unmount, snapshot create/restore + remount works
- `ROX` path now passes:
  - same `ROX` volume mounted into two sandboxes successfully
  - both sandboxes read the same seeded file
  - sandbox write attempt hits `Read-only file system`
  - file is not created
- `RWX` sandbox mount is rejected as expected
- `ROX` volume mounted onto a read-write template path is rejected as expected
- deleting a mounted `RWO` volume is rejected as expected

Note: I also started a focused remote `go test ./tests/e2e/scenarios/single-cluster ... --ginkgo.focus="sandbox volumes"` run, but it sat without emitting progress or consuming CPU meaningfully, so I stopped it instead of leaving a hung remote session behind. The new e2e coverage is in the PR and CI can exercise it there.

PR #267, opened 2026-04-22
feat: hand off RWO volume owners to bind ctlds

### LINKED ISSUE #266: RWO volume owner handoff: migrate HTTP-owned ctld volumes to bind node without dropping HTTP requests
## Summary

We want a single volume data-plane implementation: ctld is the only RWO owner, and storage-proxy only authenticates, routes, and coordinates. That removes the old storage-proxy direct-engine HTTP path, but it creates a new control-plane requirement:

- an HTTP-only volume may already be owned by ctld on node A;
- later, a sandbox pod may be claimed onto node B and require a portal bind for the same RWO volume.

In that case, the system must transfer ownership from ctld A to ctld B automatically.

The current behavior is not sufficient:

- portal bind preparation rejects active mounts for RWO volumes;
- ctld bind rejects a remote ctld owner as a conflict;
- there is no transfer state, draining protocol, or request gate for HTTP traffic during the move.

## Goal

Implement production-grade RWO owner handoff for ctld-owned volumes so that:

1. `ctld` remains the only volume data-plane owner for RWO volumes.
2. HTTP volume traffic does not fail during owner migration.
3. Portal bind can force ownership to move to the target sandbox node.
4. The handoff is race-safe under concurrent HTTP file mutations and bind attempts.

### Single data plane

- `ctld` is the only holder of `s0fs` / `fsserver` / node-local WAL for RWO volumes.
- `storage-proxy` does not hold a fallback `s0fs` engine for HTTP volume files.
- HTTP volume requests route to the current ctld owner.

### RWO handoff semantics

When a sandbox claim requires binding a volume on a different node than the current HTTP owner:

1. manager/storage-proxy initiates handoff during `prepareVolumePortalBind`.
2. target ctld pre-opens the volume from the latest committed head.
3. source ctld enters transfer state and stops accepting **new** work for that volume, but existing in-flight HTTP work is allowed to finish.
4. storage-proxy gates new HTTP requests for that volume and waits for transfer completion instead of failing them.
5. source ctld fsyncs local WAL state as required and materializes the latest committed head.
6. owner lease/epoch is transferred to the target ctld.
7. target ctld activates ownership, completes portal bind, and waiting HTTP requests are routed there.

### Failure behavior

- if transfer fails before ownership changes, source ownership remains active and waiting HTTP requests continue normally.
- stale source ctld must not be able to resume commits after target activation.

## Expected control-plane pieces

- explicit per-volume transfer state
- owner epoch / fencing token
- bind-time handoff token or comparable authorization for target ctld
- per-volume HTTP request gate in storage-proxy while transfer is in progress
- source/target ctld RPCs for prepare, freeze, finalize, and abort

## Test plan

This work must add strong coverage across layers.

### Unit tests

- transfer state transitions: idle -> preparing -> frozen -> active(new owner) -> abort
- stale epoch / stale owner commit rejection
- source ctld allows in-flight work to finish but blocks new work once frozen
- storage-proxy request gate waits and resumes instead of returning errors
- duplicate / concurrent bind requests for the same volume converge safely
- concurrent HTTP mutation + handoff + retry does not produce split-brain owner state

### Integration tests

- source ctld with active HTTP writes hands off to target ctld successfully
- target ctld opens latest committed state after source materialize
- failed target activation leaves source owner intact
- repeated transfer attempts are idempotent

### E2E / remote-kind

- HTTP write/read on an unmounted RWO volume
- while HTTP traffic is active, claim sandbox on another node and bind the volume
- verify bind succeeds, sandbox sees the latest data, and subsequent HTTP requests continue succeeding
- verify no request loss during handoff window
- verify old owner no longer serves the volume after transfer
- verify RWO conflicts still hold for simultaneous mounted ownership

## Relation to #261

This is a follow-on control-plane issue under the `s0fs v2` direction tracked in #261. The data-plane simplification from #261 requires owner migration to be first-class for production RWO volumes.

laotoutou:
## Summary
- route RWO HTTP volume access through ctld owner-only attachments instead of relying on storage-proxy direct mounts when no owner exists
- add portal-bind handoff coordination so an HTTP-owned ctld volume can move to the target bind node while new HTTP requests wait instead of failing
- add owner-only idle cleanup plus tests covering handoff wait, completion, abort, and ctld attach behavior

## Details
- add ctld owner attach API and reuse owner-only ctld sessions when a same-node bind converts an HTTP-owned volume into a portal-backed mount
- persist handoff state in storage-proxy, gate new HTTP file requests on that handoff row, and transfer mount ownership to the target ctld during bind
- keep the source ctld draining in-flight requests, materialize once, then complete or abort the transfer explicitly
- retain legacy direct-path code as fallback, but RWO HTTP requests now prefer ctld ownership when resolver/config are available

## Testing
- `go test ./ctld/internal/ctld/portal ./ctld/cmd/ctld ./storage-proxy/pkg/http ./manager/pkg/service`
- `go test ./storage-proxy/pkg/db ./ctld/internal/ctld/server`

Closes #266

### DIFF
--- a/ctld/cmd/ctld/main.go
+++ b/ctld/cmd/ctld/main.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"os/signal"
 	"strconv"
+	"strings"
 	"syscall"
 	"time"
 
@@ -243,7 +244,7 @@ func (c combinedController) UnbindVolumePortal(r *http.Request, req ctldapi.Unbi
 	}
 	resp, err := c.Portal.Unbind(r.Context(), req)
 	if err != nil {
-		return ctldapi.UnbindVolumePortalResponse{Error: err.Error()}, http.StatusBadRequest
+		return ctldapi.UnbindVolumePortalResponse{Error: err.Error()}, volumePortalErrorStatus(err)
 	}
 	return resp, http.StatusOK
 }
@@ -254,7 +255,7 @@ func (c combinedController) AttachVolumeOwner(r *http.Request, req ctldapi.Attac
 	}
 	resp, err := c.Portal.AttachOwner(r.Context(), req)
 	if err != nil {
-		return ctldapi.AttachVolumeOwnerResponse{Error: err.Error()}, http.StatusBadRequest
+		return ctldapi.AttachVolumeOwnerResponse{Error: err.Error()}, volumePortalErrorStatus(err)
 	}
 	return resp, http.StatusOK
 }
@@ -265,7 +266,7 @@ func (c combinedController) PrepareVolumePortalHandoff(r *http.Request, req ctld
 	}
 	resp, err := c.Portal.PrepareHandoff(r.Context(), req)
 	if err != nil {
-		return ctldapi.PrepareVolumePortalHandoffResponse{Error: err.Error()}, http.StatusBadRequest
+		return ctldapi.PrepareVolumePortalHandoffResponse{Error: err.Error()}, volumePortalErrorStatus(err)
 	}
 	return resp, http.StatusOK
 }
@@ -276,7 +277,7 @@ func (c combinedController) CompleteVolumePortalHandoff(r *http.Request, req ctl
 	}
 	resp, err := c.Portal.CompleteHandoff(r.Context(), req)
 	if err != nil {
-		return ctldapi.CompleteVolumePortalHandoffResponse{Error: err.Error()}, http.StatusBadRequest
+		return ctldapi.CompleteVolumePortalHandoffResponse{Error: err.Error()}, volumePortalErrorStatus(err)
 	}
 	return resp, http.StatusOK
 }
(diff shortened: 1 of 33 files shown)


laotoutou:
Follow-up on the PR E2E timeout:

Root causes from run 24770020848 / job 72473763852:
1. `Scenario single-cluster/volumes` timed out waiting for `Sandbox0Infra` Ready because the previous scenario cleanup only deleted the manifest and immediately moved on. Manager-owned template namespaces could still be terminating, so the next scenario sometimes started while `tpl-default` was in the middle of teardown/recreate.
2. `creates volumes and snapshots` still assumed a direct HTTP file write leaves the volume unmounted. That is no longer true after the ctld-owner routing changes: the write path can leave the volume ctld-mounted, and snapshot/restore correctly returns `409`.

Fixes in `a6ad2fb`:
- wait for manager-owned namespaces (`app.kubernetes.io/managed-by=sandbox0-manager`) to disappear during e2e scenario cleanup before uninstall / next scenario startup
- update the volume lifecycle case to assert `409` for snapshot on the directly-mounted volume, and run snapshot/restore lifecycle on a fresh unmounted volume

Validation:
- local: `go test ./pkg/framework ./tests/e2e/cases`
- remote kind: synced `/root/infra` to `a6ad2fb` and reran `single-cluster` against the existing kind cluster. The suite progressed through fullmode and into later scenario work without reproducing the old immediate operator-ready stall. I stopped the manual full run once it had already moved past the earlier failure point so CI could do the full verification.

PR #269, opened 2026-04-22
fix: return conflict for ctld bind portal errors

laotoutou:
## Summary
- return conflict status for ctld bind portal errors using the shared volume portal status mapper
- add a ctld HTTP test covering bind conflicts from active owners
- keep this PR scoped to the deterministic arm64  failure

## Testing
- go test ./ctld/cmd/ctld

### DIFF
--- a/ctld/cmd/ctld/main.go
+++ b/ctld/cmd/ctld/main.go
@@ -304,6 +304,7 @@ func volumePortalErrorStatus(err error) int {
 	switch {
 	case strings.Contains(message, "already has an active owner"),
 		strings.Contains(message, "actively bound to a portal"),
+		strings.Contains(message, "already bound to"),
 		strings.Contains(message, "handoff already in progress"):
 		return http.StatusConflict
 	default:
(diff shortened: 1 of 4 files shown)


laotoutou:
Follow-up for the failing PR E2E run https://github.com/sandbox0-ai/sandbox0/actions/runs/24785819856/job/72529382821

Root cause was still split across two layers:

1. `ctld` bind conflict classification was incomplete. Bind errors like `volume <id> is already bound to <path>` were still mapped to `400`, not `409`.
2. `manager` was not converting ctld bind `409` responses into `ErrClaimConflict`, so the sandbox claim API still surfaced `500` instead of `409`.

Pushed `bdd8475` with both fixes:
- treat `already bound to` as conflict in `ctld` portal HTTP status mapping
- map ctld bind conflicts to `ErrClaimConflict` in manager bind flow
- add unit tests for both paths

Verified locally:
- `go test ./ctld/cmd/ctld`
- `go test ./manager/pkg/service`

PR #358, opened 2026-05-24
Add generic function artifacts and runtime hooks

laotoutou:
## Summary

- add generic storage artifacts with API support for create/list/get/delete and materializing artifacts into sandbox volumes
- add bulk archive upload for sandbox volume files, including streaming/chunked import for large Next/OpenNext-style file sets
- add generic function runtime hooks and optional readiness bypass so platform-specific warm processes can initialize mounted artifacts without embedding provider-specific logic in sandbox0

## Validation

- `git diff --check`
- `make manifests`
- `go test $(go list ./... | grep -v '^github.com/sandbox0-ai/sandbox0/tests/e2e/')`

## Remote validation plan

After CI starts, validate the three generic primitives in the remote kind environment:

- create/list/get/delete artifact API through the gateway path
- upload a tar archive into a volume and verify extracted contents
- deploy a function revision with artifact mounts and post-claim hook behavior


laotoutou:
Abandoned per request.

PR #361, opened 2026-05-24
Optimize volume archive import path resolution

laotoutou:
## Summary
- cache resolved directory inodes while importing tar archives
- write archive files by parent inode instead of re-resolving full paths for every entry
- add coverage for repeated files under the same archive directory

## Tests
- go test ./storage-proxy/pkg/http

## Benchmark plan
- Run remote kind benchmark against an OpenNext-style archive while PR CI runs.

### DIFF
--- a/storage-proxy/pkg/http/handlers_volume_files.go
+++ b/storage-proxy/pkg/http/handlers_volume_files.go
@@ -875,6 +875,7 @@ func (s *Server) importVolumeTarArchive(ctx context.Context, volumeID, basePath 
 		return nil, false, err
 	}
 	tarReader := tar.NewReader(reader)
+	importer := newVolumeArchiveImporter(s, volumeID)
 	result := &volumeFileArchiveImportResponse{}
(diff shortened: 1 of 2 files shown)


laotoutou:
Remote benchmark on the persistent kind environment after deploying this PR:

- Artifact: OpenNext smoke app staged by s0-opennext
- Tar size: 24,995,840 bytes
- Files: 1,277
- Directories: 341
- API: PUT /api/v1/sandboxvolumes/{id}/files/archive?path=/ via cluster-gateway NodePort 30080
- Results:
  - run 1: HTTP 200, 8.609664s
  - run 2: HTTP 200, 7.976385s
  - run 3: HTTP 200, 7.815972s
- storage-proxy restarts: 0 before, 0 after

This is the same artifact shape that previously timed out/restarted storage-proxy when imported as an archive. The new path keeps the API contract unchanged and avoids the repeated directory resolution hot path.

PR #363, opened 2026-05-25
Remove Sandbox0 Functions

laotoutou:
Removes the Sandbox0 Functions capability from the core repository: function gateway, gateway APIs, runtime proxying, metering/quota dimensions, infra-operator service config, docs, OpenAPI, generated apispec code, and e2e helpers.

Validation:
- make apispec
- make manifests
- go test ./infra-operator/internal/plan ./infra-operator/internal/ownership ./manager/pkg/namespacepolicy ./pkg/metering ./regional-gateway/pkg/http ./tests/e2e/cases
- GOTOOLCHAIN=go1.25.0+auto GOWORK=off go test $(go list ./... | grep -v '/tests/e2e/scenarios/single-cluster$')\n\nNote: full go test ./... reaches Kind single-cluster E2E and is blocked locally because sandbox0ai/infra:latest is not present.


mvgijssel:
@laotoutou Hahaha just wanted to do a POC with functions and was surprised I suddenly couldn't find it. Is this feature coming back?

laotoutou:
> @laotoutou Hahaha just wanted to do a POC with functions and was surprised I suddenly couldn't find it. Is this feature coming back?

@mvgijssel I removed this feature because I haven't fully figured out the unique value and use cases of the sandbox0 function yet. Of course, I want the sandbox0 function to come back, and I would really appreciate it if you could share your use cases.

mvgijssel:
Hi sure! I'm still in the exploratory phase but wanted to create a function which:

- starts AI sandbox
- builds a new sandbox image
- rotates the pool with the new image

This is because the image contains multiple git repositories which I always want to be on the "latest" commit when folks get a sandbox from the pool. I was thinking of configuring GitHub with a webhook to call the sandbox0 function to update the sandbox image.

laotoutou:
@mvgijssel What you’ve proposed is a very valuable use case. However, using a custom image to save a Git repository might be too frequent and a bit costly. You could try saving the Git repository to a volume instead. Currently, you can create a sandbox with a volume mounted, set up a sandbox service inside to receive webhooks, and then update the Git repository in the volume. This volume can be snapshotted and also mounted by other sandboxes.

In sandbox v0.5.0, the public URL of a sandbox service has been decoupled from the sandbox pod name. The public URL can spin up a sandbox that has already been cleaned up.

mvgijssel:
Ah great idea! So 

- put all the git repos in a volume (or multiple volumes) 
- snapshot that volume
- when creating a sandbox from the pool fork the snapshot

This prevents having to store the copies in the OCI registry and from having to rotate the pool. Clever! I’ll try this

PR #375, opened 2026-05-27
Fix netd CNI chain priority

### LINKED ISSUE #373: netd TPROXY MITM is broken on several common Kubernetes distributions / CNIs
## Summary

We're evaluating sandbox0 v0.4.1 across a matrix of local Kubernetes setups and finding that netd's TPROXY-based egress interception silently fails on more configurations than the docs suggest. One root cause (chain-order vs Calico / kube-router / Canal) is fully RCA'd with a one-line code fix; a second (suspected Cilium BPF intercept) needs maintainer input. Filing one issue because the symptoms manifest differently per CNI but the diagnostic protocol is the same.

## How we test ("did netd actually intercept?")

`curl https://github.com` returning HTTP 200 from inside a sandbox is **not** sufficient evidence that netd is enforcing — for an allowed destination, netd L4-splices the connection and the pod sees the real upstream's cert. We had several false positives until we settled on this protocol:

1. Register a template with `network.mode: block-all` plus a single `trafficRules` allow for `github.com:443`.
2. Claim a sandbox from that template via `POST /api/v1/sandboxes` (the pool pre-warm pod is **not** registered with netd until claimed — `sandbox_ips` in netd logs goes from 0 → 1 on claim).
3. From inside the claimed sandbox:
   - `curl -m 8 https://github.com` should return 200.
   - `curl -m 8 https://example.com` should **fail at TLS** with `SSL_ERROR_SYSCALL` (netd closes the connection after the SNI/Host lookup denies the destination).
4. Confirm in netd logs: `"TCP decision denied" ... "host":"example.com" "reason":"l4_denied" "action":"deny"`. If you see this log line, netd is working. If example.com instead returns 200, netd was bypassed.

## Compatibility matrix (sandbox0 0.4.1)

| Host VM | K8s distro | CNI | netd MITM works? | Notes |
|---|---|---|---|---|
| OrbStack | kind (kubeadm) | kindnet | ✅ yes | docs-spec config: `kindest/node:v1.35.0` + `enable-aggregator-routing=true`, NodePort port mappings |
| OrbStack | vind (vcluster Docker driver) | Cilium 1.16 | ❌ no | TPROXY rule counter increments, sandbox traffic still reaches real upstream (HTTP 200 to denied destinations). Suspected Cilium BPF veth-ingress intercept |
| OrbStack | vind | Flannel (vcluster default) | ❌ no | TCP SYN dropped — different symptom from Cilium but same diagnosis class (TPROXY redirect doesn't deliver to listener) |
| Lima VM | kind (inside the lima VM) | kindnet | ✅ yes | Kernel-level TPROXY verified with a minimal socat+IP_TRANSPARENT test from a pod |
| Lima VM | kubeadm (direct) | flannel | ❌ no | |
| Lima VM | k3s | flannel + kube-router | ❌ no | curl times out; kube-router installs `ACCEPT ctstate RELATED,ESTABLISHED` in mangle PREROUTING before NETD_PREROUTING |
| Lima VM | RKE2 | Canal (Calico + Flannel) | ⚠️ works with workaround | Same chain-order issue as k3s; we run a 30-line systemd unit that re-pins NETD_PREROUTING to position 1 every 2s. Felix overwrites it on its sync cycle |

## Root cause #1: chain ordering in mangle PREROUTING (Calico / Canal / kube-router)

`netd/pkg/redirect/manager.go:141` attaches NETD_PREROUTING via `Append`:

```go
m.ipt.Append(mangleTable, "PREROUTING", "-j", chainName)
```

CNIs that enforce NetworkPolicy via iptables install their own chain at mangle PREROUTING position 1, and that chain starts with an `ACCEPT ctstate RELATED,ESTABLISHED` rule. Because the nat / mangle table exits processing on ACCEPT, every packet on an ESTABLISHED TCP stream short-circuits past NETD_PREROUTING. The initial SYN (ctstate NEW) does TPROXY, but the connection stalls at the next data segment — symptom is "connects, hangs at TLS".

Confirmed CNIs in this class: Calico, Calico-via-Canal (RKE2 default), kube-router (k3s default).

`netd/pkg/redirect/manager.go:159` already uses `Insert(1)` for the nat table — the mangle path is the bug. Suggested fix:

```go
m.ipt.Insert(mangleTable, "PREROUTING", 1, "-j", chainName)
```

with a corresponding update to the existence check (line 134) to also verify position.

Without this fix, sandbox0 MITM is effectively unreliable on any cluster running a NetworkPolicy-enforcing CNI in iptables mode.

### Reproducer (RKE2 / Calico)

```
iptables -t mangle -L PREROUTING -nvx
# Observe: cali-PREROUTING at position 1, NETD_PREROUTING at the tail.
iptables -t mangle -I PREROUTING 1 -j NETD_PREROUTING
# Now curl https://<allowed-host> from a fresh sandbox returns 200 in <1s.
# Felix re-inserts cali-PREROUTING within ~10s and the workaround is lost.
```

## Root cause #2 (suspected): Cilium BPF veth ingress pre-empts netfilter TPROXY

On OrbStack + vind + Cilium 1.16 (tested with `kubeProxyReplacement=false`, `routing-mode=tunnel`, `Host: Legacy` per `cilium-dbg status`):

- NETD_PREROUTING TPROXY counter increments correctly per outbound :443 SYN from the claimed sandbox.
- netd listens on `0.0.0.0:18443` with the expected caps (`CAP_NET_ADMIN`, `CAP_NET_RAW`, privileged container).
- No connections ever land on netd's listener; tcpdump on the host's `lo` using port 18443 is empty during traffic. (Note: tcpdump-by-port-18443 is not a definitive *negative* signal if netd uses `IP_TRANSPARENT` — the kernel preserves the original 5-tuple, so the right filter is by the pod's source IP. Mentioning because we initially misdiagnosed on this.)
- The pod's veth peer (`lxc<hash>`) has a Cilium `cil_from_container` tcx/ingress BPF program attached. We suspect this program is redirecting packets via `bpf_redirect` to the egress path before the kernel's marked-fwmark → local-input delivery can complete.

This is a *suspicion*, not a confirmed RCA. Filing it because it consistently breaks under multiple test protocols. If maintainers have prior experience with Cilium-versus-iptables-TPROXY conflicts, that context would help.

## Versions

- sandbox0: 0.4.1 (charts: vendored `sandbox0-infra-operator-0.4.1.tgz`)
- Kubernetes: 1.34 / 1.35 (no behavioral difference observed)
- Cilium: 1.16.4 (Helm chart, `kubeProxyReplacement=false`)
- vcluster: 0.34.0 (Docker driver / "vind" mode)
- kind: per `kindest/node:v1.35.0`
- Host kernel: OrbStack `6.19.13-orbstack-gbd1dc07b8cf4` (Linux 6.19, aarch64); Lima Ubuntu 26.04 (kernels 5.15 / 6.8 / 6.17 / 7.0 all tested, no difference)
- `CONFIG_NETFILTER_XT_TARGET_TPROXY=y` confirmed in /proc/config.gz on OrbStack kernel

## Asks

1. Patch `netd/pkg/redirect/manager.go:141` to `Insert(1)` for the mangle table to fix the Calico / kube-router / Canal class.
2. Document the Cilium BPF interaction (or fix it). Cilium is a popular CNI; "use kindnet" isn't a viable production answer.
3. Consider documenting the block-all-template test protocol as the canonical "is netd really working?" check, since several of us hit false positives with HTTP-200-on-allow-all-template.

## Workaround we ship for one class

```
# /usr/local/sbin/netd-rule-pin.sh
while true; do
  if iptables -t mangle -C PREROUTING -j NETD_PREROUTING 2>/dev/null; then
    line=$(iptables -t mangle -L PREROUTING --line-numbers \
           | awk '$2=="NETD_PREROUTING"{print $1; exit}')
    if [ -n "$line" ] && [ "$line" != "1" ]; then
      iptables -t mangle -D PREROUTING -j NETD_PREROUTING
      iptables -t mangle -I PREROUTING 1 -j NETD_PREROUTING
    fi
  fi
  sleep 2
done
```

Wrapped in a systemd unit on every node. Not production-grade — Felix wins the race occasionally.

laotoutou:
## Summary
- Keep `NETD_PREROUTING` and `NETD_NAT_PREROUTING` pinned at the top of host `PREROUTING` chains without a delete-before-insert gap.
- Add unit coverage for insert-before-cleanup behavior and iptables rule parsing.
- Add a shared netd transparent TCP egress e2e case and a CNI matrix workflow for kindnet, flannel, canal, and Cilium native + host legacy routing.

## CNI support boundary
- Supported: kindnet, flannel, canal, and Cilium `routingMode=native` with `bpf.hostLegacyRouting=true`.
- Not supported by this backend: Cilium default tunnel/VXLAN datapath, because sandbox pod traffic does not reliably traverse host iptables `mangle/PREROUTING` before Cilium's eBPF datapath forwards it.

Fixes #373

## Tests
- `go test ./netd/pkg/redirect`
- `go test ./tests/e2e/cases`
- `go test -run TestSelectScenarioManifestPaths ./tests/e2e/scenarios/single-cluster`
- `go test ./tests/e2e/scenarios/single-cluster -run TestSingleCluster -ginkgo.dry-run -ginkgo.focus="API network policy mode.*enforces transparent TCP egress through netd"`
- `bash -n tests/e2e/cni/install.sh`
- `git diff --check`

### DIFF
--- a/.github/workflows/netd-cni-e2e.yml
+++ b/.github/workflows/netd-cni-e2e.yml
@@ -1,0 +1,98 @@
+name: Netd CNI E2E
+
+on:
+  pull_request:
+    paths:
+      - ".github/actions/**"
+      - ".github/workflows/netd-cni-e2e.yml"
+      - "Dockerfile"
+      - "Dockerfile.local"
+      - "Makefile"
+      - "manager/**"
+      - "netd/**"
+      - "pkg/**"
+      - "tests/e2e/**"
+      - "go.mod"
+      - "go.sum"
+  workflow_dispatch:
+
+concurrency:
+  group: netd-cni-e2e-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  netd-cni:
+    name: netd (${{ matrix.cni }})
+    runs-on: ubuntu-latest
+    timeout-minutes: 50
+    permissions:
+      contents: read
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - cni: kindnet
+            short: kindnet
+            kind_config: tests/e2e/kind-config.yaml
+            install_cni: "false"
+          - cni: flannel
+            short: flannel
+            kind_config: tests/e2e/kind-config-no-cni.yaml
+            install_cni: "true"
+          - cni: canal
+            short: canal
+            kind_config: tests/e2e/kind-config-no-cni.yaml
+            install_cni: "true"
+          - cni: cilium-native-host-legacy
+            short: cilium
+            kind_config: tests/e2e/kind-config-no-cni.yaml
+            install_cni: "true"
+    env:
+      KIND_CLUSTER_NAME: sandbox0-netd-${{ matrix.short }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Prepare infra environment
+        uses: ./.github/actions/prepare-infra
+        with:
+          infra-dir: .
+          kind-config: ${{ matrix.kind_config }}
+          kind-cluster-name: ${{ env.KIND_CLUSTER_NAME }}
+          cache-key-suffix: netd-cni-${{ matrix.short }}
+
+      - name: Install CNI
+        if: matrix.install_cni == 'true'
+        shell: bash
+        run: bash tests/e2e/cni/install.sh "${{ matrix.cni }}"
+
+      - name: Run netd CNI E2E
+        env:
+          E2E_USE_EXISTING_CLUSTER: "true"
+          E2E_CLUSTER_NAME: ${{ env.KIND_CLUSTER_NAME }}
+        run: make test-e2e-netd-cni
+
+      - name: Dump cluster state on failure
+        if: failure()
+        shell: bash
+        run: |
+          echo "=== nodes ==="
+          kubectl get nodes -o wide || true
+          echo "=== pods ==="
+          kubectl get pods -A -o wide || true
+          echo "=== netd logs ==="
+          kubectl -n sandbox0-system logs -l app.kubernetes.io/name=netd --all-containers --tail=-1 || true
+          echo "=== sandbox0-system events ==="
+          kubectl -n sandbox0-system get events --sort-by=.lastTimestamp || true
+          kind export logs "${RUNNER_TEMP}/kind-logs-${{ matrix.short }}" --name "${KIND_CLUSTER_NAME}" || true
+
+      - name: Upload kind logs
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: netd-cni-e2e-kind-logs-${{ matrix.short }}
+          path: ${{ runner.temp }}/kind-logs-${{ matrix.short }}
+          if-no-files-found: ignore
+
+      - name: Cleanup Kind cluster
+        if: always()
+        run: kind delete cluster --name "${KIND_CLUSTER_NAME}" || true
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: all build test test-all test-integration test-integration-verbose test-e2e test-e2e-kind test-e2e-destroy test-e2e-load-images test-e2e-prepare-kind test-e2e-specific lint tidy vendor clean helm-update helm-configs release docker-build docker-build-local build-local-all docker-push proto manifests apispec oapi-codegen
+.PHONY: all build test test-all test-integration test-integration-verbose test-e2e test-e2e-kind test-e2e-destroy test-e2e-load-images test-e2e-prepare-kind test-e2e-specific test-e2e-netd-cni lint tidy vendor clean helm-update helm-configs release docker-build docker-build-local build-local-all docker-push proto manifests apispec oapi-codegen
 
 # Tool Binaries
 LOCALBIN ?= $(shell pwd)/bin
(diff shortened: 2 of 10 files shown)


laotoutou:
Adding the netd CNI compatibility notes for this PR.

The main change is that netd still uses iptables TPROXY as the primary backend, but it no longer assumes that every CNI can use the same generic TCP TPROXY path. TCP interception now follows the datapath shape of each CNI:

1. kindnet / flannel

These are bridge-based CNIs. In this path, generic TCP TPROXY can match packets in bridge netfilter but does not reliably deliver them to the local proxy. For these CNIs, TCP uses the `nat/PREROUTING -> NETD_NAT_PREROUTING -> REDIRECT` fallback. netd still recovers the original destination with `SO_ORIGINAL_DST` and applies the same policy decision before forwarding.

UDP still uses mangle TPROXY because there is no equivalent reliable REDIRECT path for preserving the original UDP destination.

2. canal / Calico

Canal uses Calico workload interfaces with the `cali+` prefix. TCP REDIRECT is not reliable here because Calico workload policy and PREROUTING chain ordering can prevent local delivery to netd. For this path, TCP uses interface-scoped TPROXY rules on `-i cali+`.

netd also pins `NETD_PREROUTING` to the top of `mangle/PREROUTING`, ahead of CNI chains such as `cali-PREROUTING`. This prevents cases where only the SYN reaches netd and later packets bypass it through CNI `RELATED,ESTABLISHED` rules. The jump update is no-gap: netd inserts the top jump first, then removes lower duplicate jumps.

3. Cilium native + host legacy routing

In this mode, Cilium pod endpoint interfaces use the `lxc+` prefix, and `bpf.hostLegacyRouting=true` makes traffic go through the host iptables path. TCP therefore uses interface-scoped TPROXY rules on `-i lxc+`. This PR validates `routingMode=native + bpf.hostLegacyRouting=true` both manually and in CI.

4. Cilium default tunnel/VXLAN

This is still outside the support boundary of the current iptables backend. With the default tunnel/VXLAN datapath, sandbox pod egress is not guaranteed to traverse host iptables `mangle/PREROUTING`, so iptables TPROXY cannot reliably intercept it. Supporting that mode would require a Cilium/eBPF-specific backend instead of extending the current iptables rules.

The new e2e coverage runs the same netd test case across the CNI matrix: kindnet, flannel, canal, and cilium-native-host-legacy. The test claims a sandbox, applies `block-all` with one allowed fixture PodIP:8080/tcp, waits for the netd applied hash, then verifies that the allowed target succeeds and the denied target fails.

PR #397, opened 2026-06-01
Fix S0FS write amplification with extent compaction

### LINKED ISSUE #396: S0FS volumes amplify disk usage ~68× on git clone
## Summary

A `git clone` (and small-write / overwrite-heavy workloads generally) into a SandboxVolume causes the S0FS backing store to grow **enormously out of proportion to the live data** — in our case ~573 MB of git working trees expanded to a **~40 GB** S0FS volume (**~68×**). The growth appears to come from un-compacted log-structured slices that are never reclaimed, so on-disk size grows ∝ *total bytes ever written*, not the live tree. This repeatedly fills the node's disk and trips kubelet `DiskPressure`.

## Environment

- sandbox0 **0.5.1** (`infra:0.5.1`), rustfs `1.0.0-alpha.79`
- Single-node cluster, S0FS object store on rustfs (S3), metadata in Postgres
- `storageProxy.config` defaults (also reproduced with `objectEncryptionEnabled: false` + `filesystemTrashDays: 0`)

## Steps to reproduce

1. Create a source SandboxVolume, mount it into a sandbox at `/workspaces`.
2. `git clone --depth 1` a handful of repositories totalling **~573 MB** of working-tree content into the volume.
3. Measure the volume's on-disk footprint in the object store (`du` of the rustfs PVC dir) and the node filesystem.

## Observed

- **~573 MB of working-tree content → ~40 GB on disk (~68×).** Stable, not transient.
- Segment objects dominate (hundreds of MB each); per-checkpoint manifests are ~12 MB and accumulate.
- A single **5 GB sequential write** (`dd`) grew the backing store by **~36 GB (~7×)** and then failed with `Input/output error`.
- Controlled micro-benchmark on a fork: writing one **~200 MB** file produced **~780 MB** of segments (~4× base); **re-writing that same file appended ~1 GB of segments per overwrite** (0.78 → 1.9 → 2.8 → 3.8 GB) while live data stayed 200 MB — i.e. superseded slices are never reclaimed.
- `du` of the live tree reports a small fraction of the real backing-store consumption (live files vs. accumulated segments).

## What I ruled out

- **Not encryption / compression.** Re-formatted a fresh fs with `objectEncryptionEnabled: false` (verified: no `s0fs.encrypted` envelope in segments) — **still ~40 GB**. git packfiles are already zlib-compressed, so lz4/encryption is irrelevant to size here.
- **Not trash retention.** `filesystemTrashDays: 0` only governs *deleted-file* retention; a clone is dominated by *new writes + index overwrites*, so it made no meaningful difference (~40 GB either way).

## Likely root cause

S0FS is log-structured (WAL + segments). `git clone` is a worst case: download packfile → repeatedly rewrite the index → check out many files. Each overwrite appends new slices to a chunk, and those superseded slices are **not compacted**, so the object store grows without bound relative to live data. (This is the same failure mode JuiceFS documents for git workloads, mitigated there by automatic threshold-based compaction + a `juicefs compact` command.)

## Expected

On-disk usage roughly proportional to the live working-tree size (a small constant factor), with slices reclaimed automatically or on demand.

## Questions / suggestions

- Is slice **compaction** implemented and triggered in 0.5.1? If so, what triggers it and why isn't it keeping up?
- Could you expose:
  - automatic compaction at a slice-count / chunk-fragmentation threshold, and/or
  - an on-demand compaction command / API, and/or
  - a CRD knob under `storageProxy.config` for compaction interval/threshold?
- A docs note that small-write / git workloads need compaction would also help.

Happy to share segment/manifest size breakdowns and `storage_proxy.config` if useful.

laotoutou:
## Summary

Fixes #396.

- replace full-file cold rewrites with extent-based write/truncate metadata so small overwrites keep unmodified cold ranges instead of re-uploading whole files
- materialize dirty data into configurable target-sized segments, defaulting to `4Mi`
- add RWO background compaction and guarded GC in both storage-proxy and ctld owners; ROX/RWX do not run background compaction because reclamation needs stronger coordination
- add `storageProxy.config` knobs for S0FS segment target size, compaction interval, dead-ratio threshold, and reclaim threshold, plus CRD/docs updates
- add unit, integration-style, and e2e regression coverage for partial rewrites, compaction, snapshot retention, access-mode behavior, and remount persistence

## Tests

- `make proto`
- `make manifests`
- `go test ./storage-proxy/pkg/s0fs ./storage-proxy/pkg/volume ./storage-proxy/pkg/snapshot ./ctld/internal/ctld/portal ./infra-operator/internal/runtimeconfig`
- `go test ./tests/e2e/cases ./tests/e2e/utils`
- `go test ./storage-proxy/...`
- `go test ./ctld/...`
- `go test ./infra-operator/...`

### DIFF
--- a/ctld/internal/ctld/portal/manager.go
+++ b/ctld/internal/ctld/portal/manager.go
@@ -2,6 +2,7 @@ package portal
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net"
 	"net/http"
@@ -362,10 +363,15 @@ func (m *Manager) Bind(ctx context.Context, req ctldapi.BindVolumePortalRequest)
 	if err != nil {
 		return ctldapi.BindVolumePortalResponse{}, err
 	}
+	segmentTargetSize, err := volume.S0FSSegmentTargetSize(m.storage)
+	if err != nil {
+		return ctldapi.BindVolumePortalResponse{}, err
+	}
 	engine, err := s0fs.Open(ctx, s0fs.Config{
-		VolumeID:    req.SandboxVolumeID,
-		WALPath:     filepath.Join(cacheDir, "engine.wal"),
-		ObjectStore: remoteStore,
+		VolumeID:          req.SandboxVolumeID,
+		WALPath:           filepath.Join(cacheDir, "engine.wal"),
+		ObjectStore:       remoteStore,
+		SegmentTargetSize: segmentTargetSize,
 		ObjectStoreForVolume: func(volumeID string) (objectstore.Store, error) {
 			return m.createObjectStore(req.TeamID, volumeID)
 		},
@@ -549,10 +555,15 @@ func (m *Manager) AttachOwner(ctx context.Context, req ctldapi.AttachVolumeOwner
 	if err != nil {
 		return ctldapi.AttachVolumeOwnerResponse{}, err
 	}
+	segmentTargetSize, err := volume.S0FSSegmentTargetSize(m.storage)
+	if err != nil {
+		return ctldapi.AttachVolumeOwnerResponse{}, err
+	}
 	engine, err := s0fs.Open(ctx, s0fs.Config{
-		VolumeID:    req.SandboxVolumeID,
-		WALPath:     filepath.Join(cacheDir, "engine.wal"),
-		ObjectStore: remoteStore,
+		VolumeID:          req.SandboxVolumeID,
+		WALPath:           filepath.Join(cacheDir, "engine.wal"),
+		ObjectStore:       remoteStore,
+		SegmentTargetSize: segmentTargetSize,
 		ObjectStoreForVolume: func(volumeID string) (objectstore.Store, error) {
 			return m.createObjectStore(req.TeamID, volumeID)
 		},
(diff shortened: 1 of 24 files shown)


laotoutou:
## Remote S0FS benchmark results

Benchmarked PR head `948de77000e30ac3889d65105a3473e32cd5fe99` on the Aliyun Singapore ECS remote test host (`go1.25.9 (Red Hat 1.25.9-1.0.1.al8)`, `linux/arm64`).

Method: a temporary file-backed object store on `/data` measured actual `segments/` and `manifests/` bytes produced by the S0FS engine. This is a remote microbenchmark, not a local e2e run.

### Large file partial-overwrite workload

Workload per row: 128 MiB file, 256 x 4 KiB overwrites (1 MiB total app overwrite bytes), `SyncMaterialize` after each overwrite, then compaction with a low fragmentation threshold and guarded GC. `Verify` means overwritten ranges still read back correctly after compaction/GC.

| Segment target | Initial amp | Fragmented amp before compact/GC | Stable amp after compact/GC | Initial segments | Fragmented segments | Stable segments | Reclaimable | Rewritten by compaction | Compact | GC | Verify |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|:---:|
| 1 MiB | 1.000x | 1.167x | 1.001x | 128 | 384 | 383 | 1.00 MiB | 112.00 MiB | 226 ms | 16 ms | yes |
| 4 MiB | 1.000x | 1.114x | 1.001x | 32 | 288 | 288 | 1.00 MiB | 127.00 MiB | 247 ms | 17 ms | yes |
| 16 MiB | 1.000x | 1.101x | 1.001x | 8 | 264 | 264 | 1.00 MiB | 127.00 MiB | 325 ms | 16 ms | yes |

Timing details:

| Segment target | Initial materialize | Total overwrite materialize | Cold reopen | Deleted segments | Deleted manifests |
|---:|---:|---:|---:|---:|---:|
| 1 MiB | 212 ms | 403 ms | 0 ms | 113 | 257 |
| 4 MiB | 217 ms | 301 ms | 0 ms | 32 | 257 |
| 16 MiB | 265 ms | 461 ms | 0 ms | 8 | 257 |

### Small-file workload

Workload: 5000 files x 4096 bytes with the 4 MiB segment target.

| Create+write | Materialize | Cold reopen | Read p50 | Read p95 | Segment count | Backing bytes |
|---:|---:|---:|---:|---:|---:|---:|
| 60 ms | 96 ms | 35 ms | 1 us | 5 us | 5 | 21.03 MiB |

### Snapshot retention check

| Snapshot read OK | Old segment retained while snapshot is retained | Old segment collectible after snapshot release | Planned segment bytes with snapshot | Planned segment bytes without snapshot |
|:---:|:---:|:---:|---:|---:|
| yes | yes | yes | 0.00 MiB | 4.00 MiB |

### Takeaways

- The benchmark does not reproduce the previous unbounded full-file rewrite amplification. After 1 MiB of cumulative partial overwrites on a 128 MiB file, fragmented backing bytes stayed around 1.10x to 1.17x including retained manifests; after compaction+GC, backing bytes returned to about 1.001x of live data.
- The 4 MiB default looks reasonable here: it reduces initial segment count from 128 (1 MiB target) to 32 while keeping stable amplification essentially identical. A 16 MiB target reduces initial segment count further to 8, but gives coarser compaction granularity.
- Snapshot retention behaved correctly: the old segment was not planned for deletion while the snapshot was retained, and became collectible once the retained-state set omitted the snapshot.

PR #403, opened 2026-06-05
Use hashed ConfigMaps for service configs

### LINKED ISSUE #402: [BUG] manager can mount stale config via subPath after an   operator-driven roll
## Summary

After an operator-driven manager roll (observed on a `0.5.1 → 0.5.2` upgrade), the `manager` pod came up running a **stale `/config/config.yaml`** that was missing the entire `credential_store:` block, even though the `manager` ConfigMap rendered by the operator contained it correctly. As a result, every credential-source create/update failed:

```
HTTP 400  failed to create credential source: encrypted_pg credential source storage is not configured
  at manager/pkg/http/handlers_credential_sources.go:67
```

Reads of credential-source *metadata* still returned `200` (metadata is unencrypted in Postgres), which masks the problem. The manager logs **no error** at startup — the store is simply never configured.

## Environment

- `sandbox0ai/infra:0.5.2` for both `infra-operator` and `manager` (no version skew).
- Single-cluster fullmode, builtin database/storage, `credential_store.default_storage_kind: encrypted_pg`.
- Deployment and ConfigMap both owned by `Sandbox0Infra/<name>`.

## Root cause

The manager mounts its config via a **ConfigMap `subPath`**:

```
volumeMounts: config -> /config/config.yaml  (subPath: config.yaml)
```

`subPath` mounts are populated once at pod creation and **never** receive subsequent ConfigMap updates. During a roll the operator updates the ConfigMap and the Deployment at essentially the same instant, with no ordering guarantee that the new ConfigMap content is committed/observable before the replacement pod’s volume is populated. So the new pod can snapshot the **pre-update** ConfigMap content.

The operator *does* stamp `infra.sandbox0.ai/config-hash` on the pod template (good — it forces a rollout on config change), but the annotation and the physically-mounted content can diverge:

- The broken pod carried `config-hash: 29e0794d…` (the hash for the config **with** `credential_store`) yet its mounted `/config/config.yaml` had **no** `credential_store:` block.
- A pod later recreated from the *same* Deployment template (ConfigMap now settled) mounted the **correct** config.

So the hash mechanism triggers the roll, but doesn’t guarantee the rolled pod sees the hashed content — and `subPath` prevents any self-healing.

## Evidence / timeline

| time (UTC) | event |
|---|---|
| 08:23:42 | operator creates secret `*-credential-source-encryption` |
| 08:23:53 | operator updates ConfigMap `*-manager` (adds `credential_store`) |
| 08:23:53 | operator (re)creates the `manager` pod |
| 10:23–10:25 | `createCredentialSource` → 400 `encrypted_pg ... not configured` |

```
# in-pod config vs ConfigMap — they differ; in-pod copy lacks credential_store:
$ kubectl exec <manager-pod> -- md5sum /config/config.yaml
4f1ad9f4bb6a52178bc6a1471249897d  /config/config.yaml
$ kubectl get cm <name>-manager -o jsonpath='{.data.config\.yaml}' | md5sum
e723562f61e913085ea5aa9e0ffa138c  -
```

The encryption secret and key were fine throughout (32-byte AES key mounted at the configured `key_file`); the only problem was the stale config.yaml.

## Reproduction (timing-dependent)

1. Cluster on infra-operator `0.5.1` with ≥1 `encrypted_pg` credential source.
2. Upgrade operator + manager to `0.5.2` (re-renders the manager config, including `credential_store`).
3. Manager pod rolls; intermittently the new pod mounts the pre-upgrade `config.yaml` (no `credential_store`) via `subPath`.
4. Any credential-source create/update → 400 `encrypted_pg ... not configured`; metadata GETs still 200.

## Expected behavior

A `manager` pod created after a config change must mount the config matching its `config-hash`. A config change must never leave the manager running stale config indefinitely.

## Suggested fixes (in order of robustness)

1. **Content-addressed ConfigMap** — render the manager ConfigMap with a name suffixed by its content hash (e.g. `…-manager-29e0794d`), mark it `immutable: true`, and have the Deployment reference that exact name. A pod referencing `…-manager-<hash>` then *cannot* mount stale content. GC superseded ConfigMaps on reconcile. This removes the race entirely and is robust regardless of `subPath`.
2. **Order the writes** — apply the ConfigMap and confirm it’s persisted/observed (its new `resourceVersion` reflected in the operator’s cache) *before* patching the Deployment template (`config-hash` + image).
3. **Drop `subPath` for the config mount** (mount the ConfigMap as a directory) so kubelet propagates updates, and/or have the manager watch & hot-reload its config. On its own this only bounds the staleness window (kubelet sync) and needs reload support, so prefer #1.

## Workaround

Delete the manager pod so the Deployment recreates it against the now-settled ConfigMap:

```
kubectl delete pod <manager-pod> -n <ns>
```

Note: `kubectl rollout restart deploy/<name>-manager` was **not** sufficient in our case — the pod was not replaced (the operator owns and reconciles the Deployment). Deleting the pod directly was required.

laotoutou:
## Summary
- create immutable, content-addressed ConfigMaps for infra-operator managed service configs
- update manager, gateways, scheduler, netd, storage-proxy, ssh-gateway, and ctld to mount the hashed ConfigMap name
- keep the config-hash pod annotation tied to the same rendered config payload and clean unused old hashed ConfigMaps after live pods stop referencing them

Fixes #402.

## Testing
- go test ./infra-operator/internal/controller/...

### DIFF
--- a/infra-operator/internal/controller/pkg/common/common.go
+++ b/infra-operator/internal/controller/pkg/common/common.go
@@ -59,6 +59,12 @@ type LocalDevConfig struct {
 
 const PodTemplateConfigHashAnnotation = "infra.sandbox0.ai/config-hash"
 
+const (
+	ServiceConfigBaseNameAnnotation = "infra.sandbox0.ai/service-config-base-name"
+	ServiceConfigHashAnnotation     = "infra.sandbox0.ai/service-config-hash"
+	serviceConfigNameHashLength     = 16
+)
+
 func NewResourceManager(client client.Client, scheme *runtime.Scheme, imagePullPolicy *corev1.PullPolicy, localDev LocalDevConfig) *ResourceManager {
 	return &ResourceManager{
 		Client:          client,
@@ -86,6 +92,17 @@ type ServiceDefinition struct {
 	ReadinessProbe     *corev1.Probe
 	ServiceAccountName string
 	Resources          *corev1.ResourceRequirements
+}
+
+type ServiceConfigRef struct {
+	ConfigMapName string
+	Hash          string
+}
+
+// PodAnnotations returns the pod-template annotations that roll workloads when
+// the referenced service config changes.
+func (r ServiceConfigRef) PodAnnotations() map[string]string {
+	return ConfigHashAnnotationFromHash(r.Hash)
 }
 
 // ReconcileDeployment creates or updates a deployment.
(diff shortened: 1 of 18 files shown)


laotoutou:
Remote manual validation on the Aliyun kind environment:

- built this branch into `sandbox0ai/infra:latest` and deployed `fullmode.yaml` with the infra-operator chart
- verified manager Deployment uses hashed immutable ConfigMap `fullmode-manager-config-e58a03aa7a0ffa43`
- verified Deployment `infra.sandbox0.ai/config-hash` matches ConfigMap `infra.sandbox0.ai/service-config-hash`
- verified manager pod `/config/config.yaml` md5 matches ConfigMap `config.yaml` md5 (`362a45a7c9555f51783262ec37651d30`)
- verified manager config contains `credential_store.default_storage_kind: encrypted_pg` inside the pod
- verified deployed service config volumes for cluster-gateway, manager, ctld, and netd point at `*-config-<hash>` ConfigMaps

`go test ./infra-operator/internal/controller/...` passed locally.

Note: the remote fullmode install stayed at 7/9 because the `dins` template image is missing `/usr/local/bin/sandbox0-dockerd-entrypoint`; the default template was Ready and the service config validation path passed. This appears unrelated to this ConfigMap rollout change.

mvgijssel:
Many thanks!

PR #431, opened 2026-06-10
Optimize sandbox pause/resume checkpoints

### LINKED ISSUE #430: Simplify sandbox pause/resume and optimize rootfs checkpoint diff
## Context

The current user-facing sandbox pause/resume API is checkpoint-based:

- `POST /api/v1/sandboxes/{id}/pause` saves the writable rootfs checkpoint through ctld, then deletes the runtime pod.
- `POST /api/v1/sandboxes/{id}/resume` creates or reuses a runtime pod, applies the latest rootfs checkpoint, binds portals, and initializes procd.
- Process state, memory, sockets, PIDs, and live REPL sessions are explicitly not preserved across pause/resume.

Given that model, the annotation-driven cgroup freeze/power-state path does not appear necessary for the pause/resume product semantics. It adds runtime-specific complexity for runc/gVisor target resolution, readiness-gate transitions, thaw-before-delete handling, resource resize side effects, and extra failure modes without preserving process state.

We are also seeing slow pause behavior. The main performance issue is likely that ctld currently asks containerd to diff the active rootfs snapshot through the generic containerd diff service. On Linux, containerd's default diff service uses the walking differ: it mounts lower/upper views and walks/comparares the filesystem tree. This is clean from an API perspective, but it does not take advantage of overlayfs' upperdir, which already represents the writable change set.

## Proposal

Simplify the sandbox pause/resume lifecycle and optimize rootfs checkpoint creation in one change set:

1. Remove the cgroup freeze based sandbox power-state mechanism from the normal pause/resume lifecycle.
2. Add a ctld rootfs diff fast path for overlayfs snapshots that creates checkpoint diffs from the active snapshot's overlay upperdir, with the current containerd `CreateDiff` path kept as a fallback.

The pause/resume source of truth should remain:

- pause: rootfs checkpoint + runtime pod deletion + durable sandbox record marked paused
- resume: new/reused runtime pod + rootfs checkpoint restore + procd initialization + durable sandbox record marked running

## Rootfs checkpoint fast path

ctld should continue to use CRI/containerd APIs for discovery and validation:

- resolve the running `procd` container
- inspect containerd metadata
- record runtime, runtime handler, snapshotter, snapshot key, parent chain, and base image metadata

For save/checkpoint creation:

- if the snapshotter is `overlayfs` and the active snapshot mount exposes a usable `upperdir`, generate an OCI-compatible diff tar from the upperdir
- correctly encode overlay whiteouts and opaque directories
- preserve file metadata needed by restore, including type, mode, uid/gid, symlinks, mtimes, and relevant xattrs/capabilities where supported
- compute digest and size for the generated diff descriptor
- upload the diff to the rootfs object store
- if the fast path is unsupported or fails with an unsupported-case error, fallback to the existing containerd `rootfs.CreateDiff` path

For restore/apply, keep using the existing containerd apply path initially. The fast path should produce a standard OCI layer diff that containerd can apply.

This should work for both runc and gVisor as long as the runtime rootfs uses the containerd overlayfs snapshotter. runc/gVisor differences should remain in runtime metadata and process isolation, not in rootfs diff semantics.

## Scope to evaluate/remove

- ctld power controller/reconciler cgroup freeze/thaw path
- manager power-state annotations and helper logic if no longer needed
- pod readiness behavior tied to `sandbox0.ai/power-state-*`
- thaw-before-delete compatibility paths, while keeping deletion robust for stale annotations from older pods
- runtime-specific freeze target discovery for runc/gVisor if it is only used by this mechanism
- ctld rootfs diff provider abstraction, for example:
  - overlay upperdir diff provider
  - existing containerd walking diff provider fallback
- checkpoint metrics/logging for diff provider, diff size, snapshot usage, and phase durations

## Acceptance criteria

- `POST /api/v1/sandboxes/{id}/pause` still checkpoints rootfs and releases the runtime pod.
- `POST /api/v1/sandboxes/{id}/resume` still restores the checkpoint and returns a usable sandbox.
- The flow works for both runc and gVisor runtime classes.
- Normal pause/resume no longer emits or depends on `sandbox0.ai/power-state-*` cgroup freeze annotations.
- For overlayfs active snapshots, ctld can create rootfs checkpoint diffs from overlay upperdir without invoking containerd's generic walking diff.
- Non-overlayfs snapshots, unsupported overlay states, or fast-path unsupported cases fallback to the existing containerd `CreateDiff` implementation.
- The fast path correctly handles create, modify, delete, rename, deleted directories, opaque directories, symlinks, permissions, ownership, and relevant xattrs/capabilities.
- The generated checkpoint diff remains a standard OCI layer that the existing restore/apply path can consume.
- Tests cover both runc and gVisor pause/resume behavior where available.
- Existing rootfs pause/resume e2e coverage continues to pass.
- Tests/docs are updated to reflect that pause/resume does not preserve process state and does not use cgroup freeze.

## Notes

Containerd API usage is still preferred for discovery, validation, and restore/apply. The proposed fast path is limited to checkpoint diff creation because containerd's default Linux diff service is intentionally generic and walks lower/upper mount views. That path is correct but too expensive for pause latency when the overlay upperdir already contains the writable change set.

laotoutou:
Closes #430.

## Summary
- remove ctld cgroup freeze/reconciler wiring from sandbox pause/resume and stop emitting power-state annotations
- keep ctld probe forwarding while resolving pods without cgroup/proc scans
- add an overlayfs snapshotter fast path that packages the active snapshot upperdir as an OCI layer diff, with fallback to containerd CreateDiff
- mount containerd data root read-only into ctld and document the checkpoint behavior

## Tests
- go test ./ctld/cmd/ctld ./ctld/internal/ctld/power ./ctld/internal/ctld/rootfs ./manager/pkg/controller ./manager/pkg/service ./infra-operator/internal/controller/services/ctld ./tests/integration/internal/tests/manager

## Follow-up validation
- PR CI owns full e2e
- I will validate pause/resume timings for runc and gVisor in the remote kind environment and comment on this PR with the results


laotoutou:
Remote pause/resume timing from the Aliyun ARM64 kind environment, running this PR image (`7e9a7b34`).

Environment:
- Sandbox0 fullmode on kind, `fullmode` Ready 9/9.
- runc: default kind/containerd runtime.
- gVisor: temporary `gvisor-rootfs` RuntimeClass using `runsc release-20260601.0`, containerd handler `io.containerd.runsc.v1`, with `--overlay2=none --file-access=shared`.
- Each sample: claim `default`, write a marker file through the file API, call `POST /api/v1/sandboxes/{id}/pause`, call `POST /api/v1/sandboxes/{id}/resume`, then read the marker back. Times below are API wall-clock durations.

| runtime | sample | pause | resume | HTTP | marker restored | marker path |
| --- | ---: | ---: | ---: | --- | --- | --- |
| runc | 1 | 1364 ms | 3987 ms | 200/200 | yes | `/tmp/pause-resume-marker.txt` |
| runc | 2 | 2214 ms | 2984 ms | 200/200 | yes | `/tmp/pause-resume-marker.txt` |
| runc | 3 | 1580 ms | 481 ms | 200/200 | yes | `/tmp/pause-resume-marker.txt` |
| gVisor | 1 | 1183 ms | 4341 ms | 200/200 | yes | `/pause-resume-marker.txt` |
| gVisor | 2 | 2374 ms | 4834 ms | 200/200 | yes | `/pause-resume-marker.txt` |
| gVisor | 3 | 2569 ms | 11983 ms | 200/200 | yes | `/pause-resume-marker.txt` |

Summary:
- runc avg: pause 1.72s, resume 2.48s.
- gVisor avg: pause 2.04s, resume 7.05s. gVisor resume had one cold-start outlier at 11.98s.

Note: I initially used `/tmp/pause-resume-marker.txt` for gVisor and saw the marker read successfully before pause but disappear after resume. Inspecting the node showed it was not present under the containerd live rootfs, so it was not part of the rootfs diff. The final gVisor validation above uses `/pause-resume-marker.txt`, which is visible in the containerd rootfs and restores correctly.

Follow-up on current PR head: the timings above were captured on commit `7e9a7b34`. The current PR head `4325f44f` only moves `github.com/containerd/continuity` from indirect to direct in `go.mod` for tidy; no implementation files changed after the timing run.

PR #435, opened 2026-06-11
fix(netd): support h2 for HTTPS credential interception

### LINKED ISSUE #433: [Feature] Support http/2 for netd credentials
## Summary

When a `credentialRule` uses `tlsMode: terminate-reoriginate`, netd terminates TLS with its MITM CA and only advertises **`http/1.1`** in the TLS ALPN extension. A client that offers **only `h2`** (HTTP/2) in its ClientHello ALPN therefore shares no protocol with the server and the handshake aborts with `tls alert 120` (`no_application_protocol`).

Clients that offer **both** `h2` and `http/1.1` are unaffected — netd selects `http/1.1` and they transparently downgrade. The break is specific to **HTTP/2-only** clients.

## Environment

- netd / infra image: `sandbox0ai/infra:0.6.0`
- Cluster: kind/vind on OrbStack (single node), Cilium CNI
- MITM CA issuer observed: `O = sandbox0, CN = sandbox0-netd-mitm-ca.minimal`

## Reproduction (minimal, no app SDK needed)

From inside a sandbox whose template has a `terminate-reoriginate` `credentialRule` for some host `$H` (any such host; here a public HTTPS host reached through the rule):

```sh
H=bedrock-runtime.us-west-2.amazonaws.com   # any terminate-reoriginate host

# HTTP/2 only  -> FAILS
echo | openssl s_client -connect $H:443 -servername $H -alpn h2
# ...
# SSL routines:ssl3_read_bytes:...:SSL alert number 120
# No ALPN negotiated

# HTTP/1.1 only -> OK
echo | openssl s_client -connect $H:443 -servername $H -alpn http/1.1
# ALPN protocol: http/1.1

# Both -> OK (netd selects http/1.1)
echo | openssl s_client -connect $H:443 -servername $H -alpn h2,http/1.1
# ALPN protocol: http/1.1
```

## Actual behavior

An `h2`-only ClientHello produces `SSL alert number 120` (`no_application_protocol`) and the handshake aborts. Certificate verification itself succeeds (the connection does reach netd's MITM), so this is purely ALPN negotiation.

## Expected behavior

A terminate-reoriginate MITM should not break HTTP/2-only clients. Ideally netd advertises and supports `h2` on the client-facing side (proxying as appropriate to the upstream); at minimum the limitation should be documented and, ideally, configurable per rule.

## Impact

HTTP/2-only clients cannot use any terminate-reoriginate host. The concrete case that surfaced this: the **AWS SDK's Bedrock streaming API** (`bedrock-converse-stream`) negotiates `h2` only, so every streaming Bedrock call through a credential-injecting rule fails with alert 120. gRPC and other h2-only clients would hit the same wall. Clients built on Node/undici, Go's default transport, browsers, curl, etc. are unaffected because they offer both protocols and fall back to `http/1.1`.

## Workaround

Force the client onto HTTP/1.1. For the AWS Bedrock case, set `AWS_BEDROCK_FORCE_HTTP1=1`. There is no workaround for clients that cannot be told to downgrade.

## Suggested fix

Add `h2` to the ALPN protocol list offered by netd's terminate-reoriginate TLS listener and support HTTP/2 on that hop (or, if h2 proxying is out of scope, document the http/1.1-only limitation and expose a knob so operators know h2-only clients are unsupported on intercepted hosts).

laotoutou:
## Summary
- advertise h2 for HTTPS terminate-reoriginate credential rules while keeping HTTP/1.1 preferred for dual-stack clients
- route negotiated h2 HTTPS requests through the existing HTTP/2 proxy path so injected headers are preserved upstream
- document that netd preserves the negotiated downstream ALPN and does not silently downgrade h2 to HTTP/1.1 upstream

Fixes #433

## Tests
- go test ./netd/pkg/proxy -count=1 -timeout=3m
- go test ./netd/pkg/policy ./netd/pkg/proxy -count=1 -timeout=3m

### DIFF
--- a/docs/credential/egress-auth/page.mdx
+++ b/docs/credential/egress-auth/page.mdx
@@ -63,6 +63,8 @@ When a credential rule uses TLS interception such as `protocol: https`, `protoco
 
 At procd startup, Sandbox0 writes a combined CA bundle containing the image's system roots plus the netd MITM CA, then exports it through common TLS environment variables such as `SSL_CERT_FILE`, `NODE_EXTRA_CA_CERTS`, `REQUESTS_CA_BUNDLE`, `CURL_CA_BUNDLE`, `GIT_SSL_CAINFO`, and `AWS_CA_BUNDLE`. Applications normally do not need to handle the MITM CA themselves.
 
+For HTTPS credential injection, netd supports both HTTP/1.1 and HTTP/2 on `terminate-reoriginate` connections. Clients that offer HTTP/2 negotiate HTTP/2; HTTP/1.1-only clients continue to use HTTP/1.1. netd preserves the negotiated downstream ALPN protocol when re-originating upstream traffic and does not silently downgrade an HTTP/2 downstream request to HTTP/1.1 upstream.
+
 ## Example Policy
 
 This policy allows access to the GitHub API, binds a token source, and injects an `Authorization` header for matching HTTPS requests:
--- a/netd/pkg/proxy/tls_proxy.go
+++ b/netd/pkg/proxy/tls_proxy.go
@@ -150,6 +150,9 @@ func downstreamTLSNextProtos(req *adapterRequest) []string {
 		}
 		return []string{"h2"}
 	}
+	if req != nil && req.EgressAuth != nil && egressAuthMayUseHTTPSH2(req.EgressAuth) {
+		return []string{"h2", "http/1.1"}
+	}
 	if req != nil && policy.RequiresProtocolTLSTermination(req.Compiled, req.Host, req.DestPort, "mcp") {
 		return []string{"h2", "http/1.1"}
 	}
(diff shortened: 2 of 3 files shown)


laotoutou:
Correction/update on validation:

I added real remote sandbox validation on the Aliyun Singapore kind environment after updating this PR:

- Synced the PR branch to the remote workspace, rebuilt `sandbox0ai/infra:latest`, loaded it into kind, and restarted both `fullmode-netd` pods.
- Confirmed both netd pods are `1/1 Running` on the two worker nodes.
- Used a real claimed sandbox pod: `rs-mrswmylvnr2a-default-x2ggl`.
- Created a unique `static_headers` credential source and a sandbox network policy with:
  - `protocol: https`
  - `tlsMode: terminate-reoriginate`
  - `domains: ["nghttp2.org"]`
  - `credentialRef: h2-curl-ref-1781175935`
- Waited until `sandbox0.ai/network-policy-hash` matched `sandbox0.ai/network-policy-applied-hash` on the sandbox pod.
- Ran the request from inside the sandbox container with HTTP/2 enabled:

```bash
CURL_CA_BUNDLE=/tmp/sandbox0/netd-ca-bundle.crt \
  curl --http2 -fsS --max-time 20 \
  -w "\nCURL_HTTP_VERSION:%{http_version}\n" \
  https://nghttp2.org/httpbin/headers
```

Result from inside the sandbox:

```text
{"headers":{"Accept":"*/*","Accept-Encoding":"gzip","Authorization":"Bearer real-curl-h2-token-1781175935","Host":"nghttp2.org","Transfer-Encoding":"chunked","User-Agent":"curl/8.5.0"}}

CURL_HTTP_VERSION:2
SANDBOX_REAL_CURL_H2_EGRESS_AUTH_OK sandbox=rs-mrswmylvnr2a-default-62pxp pod=rs-mrswmylvnr2a-default-x2ggl ref=h2-curl-ref-1781175935 token=real-curl-h2-token-1781175935 http_version=2
```

I also used a temporary h2-only Go client inside the sandbox (`x/net/http2.Transport`, ALPN only `h2`) before the ALPN ordering adjustment. That verified h2-only clients succeed through the real sandbox egress path and receive the injected header.

One behavior found during the real test: with ALPN ordered as `http/1.1,h2`, `curl --http2` still negotiated HTTP/1.1. I updated the PR so HTTPS credential interception advertises `h2,http/1.1`; HTTP/1.1-only clients still negotiate HTTP/1.1, while clients offering HTTP/2 negotiate HTTP/2.

Cleanup: cleared the sandbox network policy, deleted the temporary credential source, and deleted the claimed sandbox.

PR #553, opened 2026-06-19
Fix optional volume mounts in rootfs persistence

laotoutou:
## Summary
- keep template-declared but unbound volume mount paths as ordinary sandbox rootfs directories
- inject user volume portals only for claim-time bound mounts and send volume-bound claims through cold runtime creation
- filter rootfs checkpoint/apply paths from actual bound mounts instead of all template-declared portals
- document optional mount/rootfs behavior

## Tests
- go test ./manager/pkg/apis/sandbox0/v1alpha1 ./manager/pkg/service ./manager/pkg/controller
- go test ./ctld/internal/ctld/rootfs

## Note
- go test ./ctld/internal/ctld/portal currently fails before tests with missing generated package github.com/sandbox0-ai/sandbox0/storage-proxy/proto/fs in this checkout.


laotoutou:
Closing at author request; this approach would block runtime binding semantics for declared optional mounts.

PR #557, opened 2026-06-20
Make template volume mounts optional by default

### LINKED ISSUE #556: Allow optional template volume mounts at claim time
## Problem

SandboxTemplate `spec.volumeMounts` declares fixed volume portal mount points. Today, if a template declares any volume mount, every declared mount point must be bound to a concrete SandboxVolume in the claim request. This prevents templates from exposing optional volume mount points.

The requested behavior is narrower than dynamic mounting: users may omit selected template-declared mounts at claim time, and Sandbox0 does not need to support binding them later.

## Design

Add an explicit `optional` flag to `VolumeMountSpec`:

```yaml
spec:
  volumeMounts:
    - name: workspace-data
      mountPath: /workspace/data
      optional: true
```

Semantics:

- `optional: false` or omitted preserves the current behavior: the claim request must bind the mount point.
- `optional: true` allows the claim request to omit that mount point.
- Claim requests may still bind only paths declared by the template.
- Omitted optional mounts still publish the CSI volume portal and remain present as FUSE mount points in the pod.
- No late-bind API is required; omitted optional mounts stay unbound for the sandbox runtime.
- Only claim-provided mounts are persisted in sandbox `mounts`, reported in `bootstrap_mounts`, and excluded from rootfs checkpoint/restore/fork capture.

## Rootfs / FUSE notes

ctld already publishes template volume portals with an unbound FUSE session before claim binding. An omitted optional mount should therefore appear as an unbound portal directory, while actual bound volume mount paths continue to be excluded from rootfs operations.

Validation should specifically cover the rootfs behavior so unbound optional mount directories do not cause volume contents to be captured or excluded incorrectly.

## Tasks

- Add `optional` to the template volume mount schema/types.
- Update claim validation to require only non-optional template mounts.
- Preserve undeclared mount rejection and access-mode checks for provided mounts.
- Update OpenAPI/docs and generated artifacts through the normal sync path.
- Add unit tests for required vs optional mounts and rootfs excluded paths.
- Validate on the remote kind environment with a template that has an omitted optional mount and a bound mount, including rootfs directory behavior.

laotoutou:
## Summary

- remove the explicit `optional` field from template `VolumeMountSpec`; all declared mount paths may be omitted by a claim
- create idle pods without user volume portals, and create cold claim pods with only the claim-bound portal subset
- initialize all declared mount point directories in procd so omitted paths are writable rootfs directories and persist through rootfs snapshots/restores
- update OpenAPI, CRDs, generated types, docs, and focused tests

Closes #556

## Tests

- `go test ./manager/pkg/apis/sandbox0/v1alpha1 ./manager/pkg/controller ./manager/pkg/service ./manager/procd/pkg/http/handlers ./pkg/apispec ./pkg/template/http`
- `go test ./manager/... ./pkg/...`

### DIFF
--- a/ctld/cmd/ctld/main.go
+++ b/ctld/cmd/ctld/main.go
@@ -129,7 +129,7 @@ func main() {
 	httpServer := newHTTPServer(httpAddr, combinedController{
 		Controller: probeController,
 		Portal:     portalManager,
-		RootFS:     buildRootFSController(storageCfg),
+		RootFS:     buildRootFSController(storageCfg, portalManager),
 	})
 	if obsProvider != nil {
 		httpServer.Handler = httpobs.ServerMiddleware(obsProvider.HTTPServerConfig(zapLogger))(httpServer.Handler)
@@ -185,7 +185,7 @@ func buildProbeController(ctx context.Context, obsProvider *observability.Provid
 	return controller
 }
 
-func buildRootFSController(storageCfg *apiconfig.StorageProxyConfig) ctldserver.RootFSController {
+func buildRootFSController(storageCfg *apiconfig.StorageProxyConfig, portalResolver ctldrootfs.PortalResolver) ctldserver.RootFSController {
 	store, err := buildRootFSObjectStore(storageCfg)
 	if err != nil {
 		log.Printf("ctld rootfs object store disabled: %v", err)
@@ -201,7 +201,8 @@ func buildRootFSController(storageCfg *apiconfig.StorageProxyConfig) ctldserver.
 			RootFSCacheDir:         filepath.Join(portalRoot, "rootfs"),
 			Namespace:              containerdNamespace,
 		}),
-		Store: store,
+		Store:          store,
+		PortalResolver: portalResolver,
 	})
 }
 
--- a/ctld/internal/ctld/portal/manager.go
+++ b/ctld/internal/ctld/portal/manager.go
@@ -64,6 +64,9 @@ type portalMount struct {
 	fs         *volumefuse.FileSystem
 	server     *fuse.Server
 
+	rootfsBackingPath string
+	rootfsSession     volumefuse.Session
+
 	volumeID  string
 	teamID    string
 	mountedAt time.Time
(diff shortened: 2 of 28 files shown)


laotoutou:
Updated the PR with the rootfs-backed unbound portal design.

Local checks:
- `go test ./ctld/internal/ctld/portal ./ctld/internal/ctld/rootfs ./ctld/cmd/ctld ./manager/pkg/apis/sandbox0/v1alpha1 ./manager/pkg/service`
- `git diff --check`

Remote validation on the Aliyun kind environment, commit `dc7b0139`:
- Deployed this PR through `/root/sandbox0ai/Makefile`; `Sandbox0Infra/fullmode` reached `Ready`.
- Created a temporary template with two declared volume portals and `minIdle=1`:
  - bound portal: `/workspace/optportal-1781957920-22292-bound`
  - unbound portal: `/workspace/optportal-1781957920-22292-rootfs`
- Verified the idle warm-pool pod had both portals pre-mounted in the pod spec and in `/proc/self/mountinfo`.
- Claimed the sandbox with only the bound portal attached to a SandboxVolume.
  - The claimed `pod_name` matched the idle pod name.
  - Pod annotation `sandbox0.ai/claim-type=hot` confirmed the claim came from the warm pool.
  - `bootstrap_mounts[0].state=mounted` for the bound volume.
- Verified the claimed and resumed pods still had both portals mounted in spec and `/proc/self/mountinfo`.
- Wrote files through the unbound portal, wrote directly under the mountpoint, and changed the unbound mountpoint directory mode to `753`.
- Paused and resumed the sandbox; the unbound portal data and mountpoint mode survived the rootfs checkpoint.
- Mutated the unbound portal to v2 and the bound volume to v2, then restored the rootfs snapshot:
  - unbound portal restored to `rootfs-v1` / `nested-v1`
  - unbound mountpoint mode restored to `753`
  - unbound portal remained writable after restore
  - bound volume stayed at `volume-v2`, proving rootfs restore excluded the bound mount path
  - direct volume API read `/volume.txt=volume-v2`
- Temporary `optportal` resources were cleaned up and the remote ECS was stopped with StopCharging.

PR checks at the time of validation: quick-check, netd matrix, and release-package jobs passed; e2e was still pending.

PR #588, opened 2026-06-23
Redesign s0fs as common persistent rootfs storage

### LINKED ISSUE #587: Redesign s0fs as the common persistent filesystem for volumes and rootfs
## Context

Sandbox0 already has a relatively mature s0fs implementation for SandboxVolume persistence:

- manifest / segment / extent state
- WAL and materialization
- committed head CAS
- object reuse across snapshots and forks
- range reads for cold segment data
- encryption, GC, and storage metering foundations

Persistent rootfs currently uses a separate tar-based checkpoint path:

- pause scans containerd overlay upperdir, serializes an OCI tar diff, and uploads it
- resume downloads tar diff layers and applies them into a claimed runtime rootfs
- DB-side rootfs state already behaves like a COW layer graph, but runtime restore is still physical tar transport and extraction

This duplicates persistent filesystem concepts and makes rootfs pause/resume too expensive. More importantly, a containerd-snapshotter-only approach would improve pause/resume while giving up Sandbox0's cold-start advantage, because resume would become coupled to container creation or snapshotter attach instead of generic idle pod claim.

## Decision

Make s0fs the bottom persistent filesystem capability in Sandbox0.

`s0fs-core` is only a working name for the new/refactored s0fs engine itself. It is not a new external component, service, daemon, snapshotter-only adapter, or second long-term storage layer. If this design is implemented, the product should still talk about the capability as s0fs.

In the target state:

- s0fs is the shared persistent filesystem engine
- SandboxVolume is one frontend over s0fs
- persistent rootfs is another frontend over s0fs
- storage-proxy owns the SandboxVolume frontend responsibilities
- ctld owns the persistent rootfs runtime and mount frontend responsibilities
- manager orchestrates lifecycle and rootfs/volume heads, but does not implement filesystem details

The current rootfs implementation does not need to be preserved as an internal compatibility layer. It can be replaced directly by the s0fs-based model as long as public OpenAPI behavior remains stable.

The compatibility boundary is the public Sandbox0 OpenAPI behavior, not the current internal rootfs save/apply implementation, manager-to-ctld contracts, or storage-proxy-private package layout.

## Non-Goals

This issue is not asking for a containerd snapshotter-only optimization that merely makes pause/resume faster while losing the existing cold-start model.

Specifically, the target design should not require:

- creating a new pod on resume
- pulling or unpacking a new image on resume
- depending on container creation as the persistent rootfs attach mechanism
- downloading and applying tar layers in the resume hot path
- writing all rootfs contents into the runtime rootfs before startup
- keeping the current OCI tar diff rootfs path as the long-term internal contract
- adding a new long-running storage component separate from s0fs, storage-proxy, ctld, and manager

## Target Architecture

Suggested conceptual ownership:

```text
pkg/s0fs
  persistent filesystem engine
  manifest / segment / extent / inode graph
  WAL / head CAS / materializer / GC / encryption
  metadata model for volume and rootfs needs

storage-proxy
  SandboxVolume frontend over s0fs
  file API / FUSE protocol / volume snapshot API

ctld
  RootFS frontend over s0fs
  rootfs attach / detach / commit / recovery
  mount namespace publish into warm sandbox pods

manager
  lifecycle orchestration
  rootfs and volume head orchestration
  no filesystem implementation details
```

The existing s0fs implementation should no longer be treated as storage-proxy-private code. It should be refactored into the shared s0fs engine used by both storage-proxy and ctld.

Volume and rootfs are frontends over the same filesystem engine, not two separate persistence systems.

## Volume Frontend

The SandboxVolume frontend remains focused on durable user data:

- explicit volume lifecycle
- POSIX-ish file operations through storage-proxy/procd protocols
- volume snapshot, fork, restore
- access mode and ownership rules
- storage metering and GC

It should use the common s0fs engine for persistent state, manifests, segments, object reuse, encryption, metering source data, and head management.

## RootFS Frontend

The persistent rootfs frontend should be provided by ctld.

The sandbox pod should remain a generic warm execution shell:

- procd runs as an outer supervisor in the prewarmed pod rootfs
- the user process rootfs is a ctld-published s0fs-backed rootfs mount
- claim/resume attaches a rootfs head to an already running idle pod before procd initializes the user runtime
- user cmd/repl/function processes enter that mounted rootfs through mount namespace / chroot / pivot-root style isolation

This preserves the existing cold-start advantage: idle pods can be prepared before a persistent rootfs head is known.

## RootFS Resume Hot Path

The intended resume path is:

1. manager reads the sandbox rootfs head
2. manager claims a generic idle pod
3. manager asks ctld to attach the s0fs rootfs head to that pod
4. ctld prepares and publishes the rootfs mount into the pod namespace
5. manager initializes procd
6. procd starts user processes inside the attached rootfs

The hot path should not require:

- creating a new pod
- pulling or unpacking a new image
- downloading tar layers
- applying tar layers
- writing all rootfs files into the container rootfs before startup

Large file content should be lazy or cache-backed so resume latency scales with manifest/mount work and first-access working set, not total rootfs byte size.

## RootFS Pause Hot Path

The intended pause path is:

1. procd quiesces user processes
2. ctld freezes or detaches the active rootfs view
3. ctld commits the current writable state into a new immutable s0fs head
4. manager advances the sandbox rootfs filesystem head
5. the pod can be deleted or returned to the warm pool

Commit should persist changed metadata and changed content segments, not reserialize the whole rootfs as tar.

## s0fs Engine Requirements

The shared s0fs engine must support both Volume and RootFS semantics.

It should include or grow support for:

- inode graph and directory tree metadata
- file mode, uid, gid, nlink, size, timestamps
- symlinks and hardlinks
- xattrs
- whiteouts and opaque directories
- fifo / char device / block device metadata where supported
- sparse file / extent representation
- segment objects and range reads
- manifests and immutable heads
- committed head CAS
- WAL / crash recovery
- materialization and local cache management
- encryption
- GC over shared object references
- storage metering source data

RootFS-specific behavior such as mount namespace publish, rootfs attach/detach, and execution isolation should live in ctld as the RootFS frontend, not in storage-proxy.

## Internal Refactor Freedom

Internal implementation can be changed directly to reach this model. In particular:

- `/api/v1/rootfs/save` and `/api/v1/rootfs/apply` do not need to keep their current internal semantics
- ctld rootfs controller/runtime interfaces can be replaced
- manager lifecycle internals can be renamed or reworked
- procd can change from "process inside restored container rootfs" to "outer supervisor that launches user processes inside an attached s0fs rootfs"
- the current OCI tar diff rootfs implementation does not need to remain as the long-term path
- existing s0fs package layout can be moved and refactored so it becomes the common filesystem engine

The public OpenAPI contract should stay stable unless a separate API change is explicitly proposed and reviewed.

## Expected Performance Properties

The target design should make:

- small rootfs resume close to idle pod claim + mount publish latency
- large rootfs resume mostly independent of total rootfs byte size
- pause proportional to changed files and changed bytes
- rootfs snapshot/fork/restore close to metadata operations
- volume snapshot/fork/restore continue to reuse s0fs object state
- persistent rootfs compatible with the existing 200ms-class cold-start goal

## Correctness Requirements

The design must preserve filesystem semantics needed by sandbox workloads, including at least:

- path and directory metadata
- ownership and permissions
- symlinks and hardlinks
- xattrs
- whiteouts and opaque directories
- sparse file behavior where relevant
- safe mount namespace isolation between sandboxes
- consistent crash recovery after ctld or storage-proxy restart
- correct exclusion/binding behavior for SandboxVolume portals and procd-owned runtime paths

## Acceptance Criteria

- s0fs is the common persistent filesystem engine used by both SandboxVolume and persistent rootfs.
- Persistent rootfs resume can use a generic prewarmed idle pod.
- Resume does not physically apply tar layers in the hot path.
- Pause commit does not generate a full OCI tar diff as the primary persistent format.
- RootFS snapshot/fork/restore are implemented as s0fs head operations.
- SandboxVolume continues to use s0fs through the storage-proxy frontend.
- Internal rootfs save/apply, ctld, manager, and package boundaries may be refactored as needed.
- Public OpenAPI behavior remains stable.
- Tests cover volume behavior, rootfs attach before procd initialization, pause/resume ordering, snapshot/restore/fork behavior, crash recovery, mount isolation, sparse files, and rootfs metadata semantics.

laotoutou:
## Summary
- Move s0fs into the shared pkg layer and extend it with xattr and special node support for POSIX rootfs fidelity.
- Add s0fs rootfs head save/apply flow in ctld, including host rootfs import, manifest restore, and FUSE-backed rootfs attach.
- Teach manager to persist s0fs rootfs heads, pass filesystem/head descriptors to ctld, and initialize procd with the mounted rootfs path.
- Run procd contexts and function runners inside the mounted rootfs while keeping the public OpenAPI unchanged.

## Testing
- go test ./pkg/s0fs ./storage-proxy/pkg/fsserver ./ctld/internal/ctld/portal ./ctld/internal/ctld/rootfs ./manager/procd/pkg/process ./manager/procd/pkg/context ./manager/procd/pkg/http/handlers ./manager/pkg/service
- go test ./ctld/... ./storage-proxy/... ./pkg/... ./manager/...
- go test -buildvcs=false $(go list -buildvcs=false ./... | grep -v '^github.com/sandbox0-ai/sandbox0/tests/e2e/scenarios/single-cluster$')
- go test ./... fails only in tests/e2e/scenarios/single-cluster because this local machine does not have kind installed.

Closes #587


laotoutou:
Final validation update for PR #588 (`efd2ccd0`):

- GitHub PR E2E passed after aligning the rootfs e2e checks with the public API context contract.
- quick-check passed.
- Netd CNI matrix passed: kindnet, flannel, canal, and cilium-native-host-legacy.
- Release package metadata/infra passed; publish-images is skipped as expected for this PR.
- Remote Aliyun/kind manual validation was completed earlier and the ECS instance has been stopped with StopCharging.

One note on the e2e adjustment: resumed s0fs rootfs is now the root used by Sandbox API contexts. Direct `kubectl exec` into the implementation pod root is not treated as an OpenAPI/public contract, so the rootfs persistence e2e now verifies writes and reads through `CreateContext(wait_until_done=true)` while preserving the same TTL pause/resume and snapshot/restore/fork flows.

PR #613, opened 2026-06-26
Add ctld rootfs object cache

laotoutou:
## Summary
- add a disposable ctld node-local rootfs object cache with digest validation, max size, min-free, max-age, and sweep interval GC knobs
- persist the last runtime node for sandbox records and use it as best-effort locality for resumed idle/cold runtime pods
- expose ctld cache config through Sandbox0Infra/operator CRDs and document self-hosted usage

## Testing
- make manifests
- make proto
- go test ./ctld/internal/ctld/rootfs ./ctld/cmd/ctld ./infra-operator/internal/controller/services/ctld ./manager/pkg/service ./tests/integration/internal/tests/manager
- GOFLAGS=-buildvcs=false go list ./... | grep -v '/tests/e2e/' | xargs go test -buildvcs=false\n\n## Notes\n- `go test ./...` was also attempted; non-e2e packages compiled after fixing the integration test store, but e2e scenarios require `kind`, which is not installed in this environment.\n


laotoutou:
Follow-up issue for configurable node affinity and rootfs cache metrics: #614

PR #626, opened 2026-06-30
Remove manager autoscaler and add cold claim backpressure

laotoutou:
## Summary

- Remove the unused manager pool autoscaler implementation and its CRD/config surface.
- Add manager-side cold claim concurrency backpressure with `coldStartConcurrency` / `cold_start_concurrency` config.
- Return retryable `503 unavailable` when cold claim capacity is saturated instead of letting requests pile into CNI startup and fail later.
- Update generated deepcopy/CRD files and docs that referenced the old autoscaler behavior.

## Context

The recent ACK/Terway benchmark showed cold claims failing primarily while waiting for pod network identity. The old manager autoscaler path was also ineffective because the sandbox service never received an autoscaler instance, so this PR removes it instead of trying to patch the old behavior.

## Verification

- `make manifests`
- `go test ./manager/pkg/service ./manager/pkg/controller ./infra-operator/api/config ./infra-operator/internal/runtimeconfig`
- `go test ./manager/... ./infra-operator/...`
- `git diff --check`


laotoutou:
Canceled per request; closing PR and deleting the branch.

PR #628, opened 2026-06-30
Improve autoscaler cold claim backpressure

laotoutou:
## Summary

- wire the operator autoscaler into `SandboxService` and let `PoolManager` preserve autoscaler-owned ReplicaSet targets within `minIdle/maxIdle`
- add synchronous cold-claim admission before `createNewPod`, using per-template in-flight limits plus observed active pods without PodIP as backpressure
- trigger proactive scale checks on hot claims and increase generated stable sandbox ID entropy without changing pod name suffix semantics

## Why

The benchmark showed cold claims creating hundreds of active pods on one node while the idle ReplicaSet stayed at `minIdle`. Most failures then timed out waiting for pod network identity because PodIP assignment lagged behind the burst. The manager should replenish the idle pool quickly, but it should not keep creating cold pods when the network-identity backlog is already saturated.

## Testing

- `go test ./manager/pkg/controller ./manager/pkg/service ./manager/pkg/http`
- `go test ./manager/...`
- `go test ./...` was attempted, but this checkout cannot complete the full suite because `storage-proxy/proto/fs` generated package is missing for storage-proxy/ctld setup, and local e2e requires `kind`, which is not installed in this environment.

### DIFF
--- a/manager/cmd/manager/main.go
+++ b/manager/cmd/manager/main.go
@@ -330,6 +330,7 @@ func main() {
 	sandboxService.SetCredentialStore(credentialStore)
 	sandboxService.SetQuotaStore(quotaRepo)
 	sandboxService.SetSandboxStore(sandboxStore)
+	sandboxService.SetAutoScaler(operator.GetAutoScaler())
 	rootFSObjectStore, rootFSObjectStoreErr := buildRootFSObjectStore(cfg)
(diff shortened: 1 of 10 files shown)


laotoutou:
Superseded by #629 after resetting and redesigning the autoscaler implementation.

Issue #674, opened 2026-07-09
CSI/FUSE volumes remain after sandbox deletion and drive kubelet CPU


laotoutou:
## Summary

High-density sandbox pods can leave CSI/FUSE volumes behind during deletion. On the ACK worker this caused kubelet to stay CPU-bound after the benchmark stopped, even though the sandbox API delete calls had already completed.

This does not look like idle sandbox workload CPU. The dominant CPU consumer is kubelet while it repeatedly handles pod termination and volume cleanup for `sandbox0` CSI volumes.

## Production evidence

Environment:

- Region endpoint: `ali-ue1.sandbox0.ai`
- Worker: `i-0xi50d84g6mt4mptcfog` / `us-east-1.10.60.6.176`
- Template: `default`
- Claim config: sequential claim, `128Mi`, `hard_ttl=3600s`
- Run ID: `ack-default-500-ttl3600-cpuprofile-20260709T041542Z`

Benchmark result:

- Started: `2026-07-09T04:17:28Z`
- Finished: `2026-07-09T04:39:43Z`
- Completed accepted attempts: `320`
- Successes: `310`
- Failures: `10`
- Failure classes: `pod_ip_not_assigned=9`, `transport_error=1`
- Stopped reason: `stopped after 8 consecutive failures`
- Success latency: `p50=0.597s`, `p90=0.715s`, `p99=26.66s`
- All accepted-attempt latency: `p50=0.602s`, `p90=0.723s`, `p99=90.62s`

CPU attribution during hold:

- Node busy average: `86.9%`
- Load1 average: `12.98`
- Average CPU by process group, in percent of one core:
  - `kubelet=461.6`
  - `containerd=16.2`
  - `runsc_sandbox=18.8`
  - `containerd_shim_runsc=8.45`
  - `runsc_gofer=1.85`
  - `ctld=6.14`
  - `netd=0.16`
  - `terwayd=0.11`

After deleting all API-visible sandboxes:

- `s0 sandbox delete` completed with `cleanup_exit=0 duration_sec=12`
- `s0 sandbox list -o json` returned `Count=0`
- K8s still showed `346` template pods, with `331` deleting
- Worker process snapshot still showed:
  - kubelet at about `509%` CPU in `top`
  - `containerd-shim-runsc-v1=654`
  - `runsc-sandbox=327`
  - `runsc-gofer=654`

Kubelet log count for a 5-minute window:

- Total kubelet lines: `136983`
[...]

PR #681, opened 2026-07-11
Add recoverable node-wide ctld FUSE shards

laotoutou:
## Summary

- replace one-process-owned FUSE portal connections with a fixed set of node-wide ctld shards while keeping normal CSI bind mounts for gVisor and runc
- persist node-local desired state and negotiated connection identities so a replacement ctld resumes the same kernel FUSE connections instead of reconstructing sandbox runtimes
- make rootfs and S0FS mutation replay exact-once across a ctld crash, and fence route switches behind reply barriers and cache-domain generation changes
- fail closed unless kernel recovery, exact resend status, reply acknowledgement, and cache-domain invalidation are all negotiated
- add capacity metrics, CRD validation, rollout documentation, and an opt-in e2e that mutates rootfs and SandboxVolume data while the owning ctld Pod is force-deleted

This is the restart-safe architectural follow-up to #674. It does not add another deployed component: ctld owns the fixed FUSE shards and stores its journal under its existing host data path.

## Architecture

- the kernel owns the long-lived FUSE connections across ctld process lifetimes
- ctld owns a durable shard/portal/binding journal and reconstructs the userspace router on startup
- each sandbox receives a bind mount of its portal subtree, so gVisor and runc continue to see a normal CSI volume
- per-portal binding generations isolate inode, handle, positive-dentry, and negative-dentry caches inside a shared connection
- new portals register an empty cache domain before route exposure, avoiding cumulative inode scans as node density rises; later bind/unbind transitions still perform full safe invalidation
- S3-backed portals are rejected in recovery-required mode until they have an equivalent exact-replay contract

## Validation

- unit/race/stress coverage for nodefs routing, journal recovery, reply barriers, rootfs replay, and S0FS replay
- `BenchmarkNodeFSJournalUpdateAtDensity`: 1000-portal durable updates are about 9.9-10.1ms on the local arm64 validation host after removing the duplicate full-journal validation pass
- all non-e2e Go packages pass; local e2e is intentionally not run by the remote-test workflow and remains owned by GitHub CI
- `go test ./ctld/... ./infra-operator/... ./pkg/volumefuse`
- `go test -race ./storage-proxy/pkg/fsserver ./storage-proxy/pkg/s0fs ./storage-proxy/pkg/volume`
- `go vet` on changed packages; `make manifests` and `make build-local-all` clean
- repository e2e case compiles and passes race/vet checks
- disposable 2C8G Aliyun ECS, patched kernel, kind, gVisor: two final forced ctld replacements during 20,000 rootfs + S0FS fsync/rename/read-back iterations; each replacement had a new UID, `restartCount=0`, and the sandbox UID stayed unchanged
- runc: two forced ctld replacements with the same recovery assertions and 20,000 verified mutations
- matched 10-round, single-threaded 4 KiB small-file comparison: NodeFS versus legacy throughput was write -1.5%, read -1.8%, list/stat -0.8%; p95 latency was write -3.6%, read -21.6%, list/stat -0.8%
- sequential density smoke with 30 claims at 128MiB / hard TTL 1000s: 30/30 succeeded, p50 6.798s, p95 11.221s, p99 11.403s, and ctld did not restart
- final cleanup validation: sandbox portal finalizer completed before volume deletion, no dynamic namespace remained, S0FS volume/head rows were both zero, and ctld logged no materialize conflict or panic

## Configuration

```yaml
spec:
    services:
        ctld:
            nodeFSShardCount: 8
            requireFUSERecovery: true
```

Enabling, disabling, or changing the shard count is a drained-node operation. The operator rejects fixed shards without mandatory recovery.

## Dependencies

- kernel: https://github.com/sandbox0-ai/linux/pull/1
- go-fuse: https://github.com/sandbox0-ai/go-fuse/pull/1


laotoutou:
Closing because this recovery design has been withdrawn.

PR #690, opened 2026-07-11
Gate sandbox audit behind enterprise licensing

laotoutou:
## Summary

- add the signed `sandbox_audit` entitlement and explicit `spec.sandboxObservability.audit.enabled` opt-in
- require and mount the enterprise license only for cluster-gateway when centralized audit is enabled
- gate both the public events query/watch API and internal events ingest API while leaving logs, runtime metrics, and the metric catalog available without the audit entitlement
- inject netd's centralized audit ingest URL only when licensed audit is configured
- update OpenAPI, generated CRDs/deepcopy code, the license signer help, and self-hosted/sandbox documentation

## Behavior and compatibility

Centralized network audit was previously enabled implicitly with the sandbox observability backend. It is now disabled unless `spec.sandboxObservability.audit.enabled` is explicitly `true`, and cluster-gateway requires a license containing `sandbox_audit` before it starts with that configuration.

Disabling audit stops new netd projection and returns `403 feature_not_licensed` from the events API. Historical logs and runtime metrics remain available, and existing ClickHouse event rows are retained until their configured TTL expires.

The centralized ClickHouse audit path remains a best-effort historical projection rather than a compliance-grade immutable ledger. Follow-up reliability work is tracked in #689.

## Tests

- `make proto manifests apispec`
- `go mod tidy` with no module diff
- `helm lint infra-operator/chart`
- `GOFLAGS=-buildvcs=false golangci-lint run ./...`
- CI-equivalent `go test -buildvcs=false -race -count=1` for all packages except `tests/e2e` and `tests/integration`
- `git diff --check`

Remote kind validation completed on the persistent Aliyun kind environment: [results](https://github.com/sandbox0-ai/sandbox0/pull/690#issuecomment-4948803275).


laotoutou:
Remote validation completed on the persistent Aliyun Singapore 3-node kind environment for head `9f3cd178`:

- Deployment reached `Ready 9/9` with the default disabled backend. Generated cluster-gateway config had `audit_enabled: false`; netd had an empty `sandbox_observability_ingest_url`.
- With builtin ClickHouse and sandbox observability enabled but audit disabled, the deployment reached `Ready 11/11`. Manager logs ingest and ctld runtime-sample ingest were configured, netd audit ingest remained empty, and cluster-gateway had no enterprise-license mount.
- Against a real claimed sandbox, audit events returned `403 feature_not_licensed` with `feature: sandbox_audit`, while historical logs returned `200`, runtime metrics returned `200`, and the metric catalog returned `200`.
- Enabling `spec.sandboxObservability.audit.enabled: true` without a license failed closed: ClickHouse and SandboxObservability were Ready, while the CR became `Degraded 10/11` with `ClusterGatewayReady=False`, reason `EnterpriseLicenseMissing`, and the missing `fullmode-enterprise-license` Secret named explicitly.
- The test sandbox was deleted and the environment was restored to the default `Ready 9/9` state; ClickHouse test resources were cleaned up.

The first image load exposed stale kind/containerd image accumulation on the persistent host. Only unused old infra images were pruned; the kind cluster and template image caches were preserved.

PR #694, opened 2026-07-12
refactor(metering): deliver ClickHouse usage through PostgreSQL outbox

laotoutou:
## Summary

- keep ClickHouse as the long-term metering truth and sole query/export backend
- capture stable ClickHouse operations in a short-lived PostgreSQL transactional outbox
- project the oldest transaction batch from manager with exact-payload retries
- move manager and storage projection state to compact PostgreSQL producer-state tables
- bootstrap active producer state from ClickHouse during upgrade without copying historical events or windows
- make manager, storage-proxy, and netd defer ClickHouse reconnection after the one-time bootstrap

## Why

The ClickHouse repository exposed `InTx` and `*Tx` methods but ignored the supplied PostgreSQL transaction. That made storage resource mutations and metering writes look atomic when they were not. A partial ClickHouse failure could commit only part of a logical batch. Manager status-only early returns also assumed the preceding projection succeeded, and netd retries recomputed the window end, producing a different logical window after a partial failure.

## Design

PostgreSQL stores only:

- pending/delivered projection operations, retained briefly for retry and dedupe
- current manager sandbox projection state
- current storage projection state and fractional byte-hour remainder
- a one-time ClickHouse state-bootstrap marker

It does not store a second historical usage ledger. Event/window history, export cursors, watermarks, and quota query projections remain in ClickHouse.

Operations captured by one PostgreSQL transaction share a batch ID. Manager claims the globally oldest pending batch, applies operations in sequence, and advances a watermark only after every older operation succeeds. ClickHouse `ReplacingMergeTree` IDs/versions make exact retries idempotent. A failed batch retries forever with bounded exponential backoff.

Expected normal visibility is about 2 seconds for lifecycle/storage mutations, the netd report interval plus projector polling (10s + ~2s by default), and up to one minute plus polling for periodic storage byte-hour windows.

## Validation

- real PostgreSQL transaction commit/rollback and stable batch retry tests
- legacy metering migration version 5 to version 6 upgrade test
- active ClickHouse projection-state bootstrap test, including stale-state protection
- manager early-return retry regression test
- netd stable failed-window/residual-window regression test
- projector partial failure and exact-payload replay tests
- race suites: manager, netd, storage-proxy, infra-operator, and `pkg/...`
- `go vet` on changed packages
- manager, storage-proxy, and netd binary builds

Remote kind validation will run while CI is active, including ClickHouse outage, PostgreSQL backlog inspection, recovery drain, and export verification.

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,7 +184,7 @@ Sandbox0 separates region-scoped control-plane services from cluster-scoped data
 | Control plane | Optional `regional-gateway`, optional `scheduler` | Tenant/API key management, cluster selection, internal routing, template distribution |
 | Data plane | `cluster-gateway`, `manager`, `ctld`, optional `storage-proxy`, optional `netd` | Sandbox lifecycle, rootfs checkpoints, process/file APIs, volume storage, network enforcement |
 | In-pod runtime | `procd` | PID 1 inside each sandbox pod, process abstraction, file I/O, volume mount operations |
-| Storage | PostgreSQL plus S3-compatible object storage | Metadata, usage truth, events, rootfs/volume data |
+| Storage | PostgreSQL, ClickHouse, and S3-compatible object storage | Transactional metadata and metering delivery, long-term usage truth, rootfs/volume data |
 
 Self-hosting is operator-first:
 
--- a/docs/self-hosted/configuration/page.mdx
+++ b/docs/self-hosted/configuration/page.mdx
@@ -72,7 +72,9 @@ Schedulers consume the manager cluster summary fields `claim_start_limit`, `clai
 
 ## Metering Boundary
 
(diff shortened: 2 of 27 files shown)


laotoutou:
Remote kind validation completed on the persistent Aliyun ECS environment for `58656f24`:

- Deployed full mode with metering and builtin ClickHouse; `Sandbox0Infra` reached `Ready 11/11`.
- Claimed and deleted real sandboxes. ClickHouse received one idempotent `sandbox.claimed`, one `sandbox.terminated`, the final runtime window, and the terminal sandbox projection state.
- Paused the operator, scaled ClickHouse to zero, and observed real netd/storage producer operations accumulating in PostgreSQL. The oldest row retried with `connection refused`; later rows stayed at zero attempts, confirming strict oldest-batch gating.
- Restarted manager, storage-proxy, and both netd pods while ClickHouse was down. All became Ready in deferred-delivery mode and continued capturing usage in PostgreSQL.
- Restored ClickHouse. The failed head succeeded on its eighth attempt, all 36 pending operations drained in order, pending/oldest-age metrics returned to zero, and replay did not duplicate the claimed event.
- Generated actual sandbox HTTPS traffic after the netd restart. ClickHouse received non-zero 10-second egress/ingress byte windows through the PostgreSQL outbox.
- Restored the operator, removed temporary sandboxes/files, and rechecked `Ready 11/11`, a healthy default template pool, and zero pending outbox rows.

A final PostgreSQL 16 `go test -race` run also passed the outbox transaction, migration, bootstrap, and exact-retry integration tests.

PR #707, opened 2026-07-15
fix(storage): preserve POSIX errno through volume portals

### LINKED ISSUE #706: S0FS reports ENOTEMPTY/EISDIR conditions as EIO through volume FUSE
## Summary

Normal POSIX filesystem conditions from S0FS are collapsed into a generic `FailedPrecondition` error and then exposed by the volume FUSE mount as `EIO` (`Input/output error`). In particular, removing a directory that still has children should return `ENOTEMPTY`, not `EIO`.

This makes ordinary concurrent filesystem activity look like volume corruption or an unavailable mount. Package managers are especially affected because they routinely race cleanup against writers and handle specific POSIX errno values.

## Observed behavior

A coding-agent workload had overlapping `npm install` / `rm -rf node_modules` processes after an interrupted command remained alive briefly. Directories gained or retained children while cleanup was in progress, which is a legitimate `ENOTEMPTY` condition.

Instead, both npm and rm received errno `-5` / `EIO`:

```text
npm error EIO: i/o error, rmdir '/workspace/node_modules/next/dist'
rm: cannot remove '...': Input/output error
```

The mounted volume remained healthy afterward: files were readable, directory traversal worked, and the sandbox/runtime had no storage provisioning error. The misleading `EIO` was therefore the user-visible bug; the concurrent non-empty directory was not evidence of volume corruption.

A minimal mounted-volume reproduction should be:

```sh
mkdir -p /workspace/non-empty/child
rmdir /workspace/non-empty
```

Expected: `ENOTEMPTY` / `Directory not empty`.

Actual: `EIO` / `Input/output error`.

## Root cause

The S0FS engine already returns the correct semantic error:

- `storage-proxy/pkg/s0fs/engine.go`: `Engine.RemoveDir` returns `s0fs.ErrNotEmpty` when the directory has children.

That information is lost at the session boundary:

- `storage-proxy/pkg/fsserver/server.go`: `mapS0FSError` maps both `ErrNotEmpty` and `ErrIsDir` to `fserror.FailedPrecondition`.
- `ctld/internal/ctld/portal/session.go`: `mapLocalS0FSError` performs the same lossy mapping for the local S0FS portal path.
- `ctld/internal/ctld/portal/rootfs_session.go`: `mapRootFSBackedError` similarly collapses `ENOTEMPTY` and `EISDIR` into `FailedPrecondition`.
- `pkg/volumefuse/fs.go`: `statusToFuse` maps every `fserror.FailedPrecondition` to `fuse.EIO`.

The HTTP volume-file path currently compensates for this missing error granularity by matching `"not empty"` in the error message, but the FUSE path cannot safely recover the original errno this way.

## Expected behavior

Preserve filesystem semantics end-to-end so that at least:

- `s0fs.ErrNotEmpty` -> `ENOTEMPTY`
- `s0fs.ErrIsDir` -> `EISDIR` (or the operation-appropriate POSIX errno)
- `s0fs.ErrNotDir` -> `ENOTDIR`
- genuine internal/storage failures -> `EIO`

Generic control-plane precondition failures should remain distinct from normal POSIX filesystem errors.

## Suggested direction

Introduce granular filesystem error codes (or an explicit POSIX errno field) across the portal/session boundary, append rather than renumber any existing wire-visible code values, and map them explicitly in `statusToFuse`.

Please cover both local and remote portal paths. Suggested regression coverage:

1. Removing a non-empty directory through a mounted S0FS volume returns `ENOTEMPTY`.
2. Directory/file type mismatches return their operation-appropriate errno rather than `EIO`.
3. Local S0FS, remote storage-proxy, and rootfs-backed portal mappings agree.
4. Genuine `FailedPrecondition` and internal errors do not get mistaken for POSIX conditions.

laotoutou:
Fixes #706

## Summary

- retain POSIX errno details alongside high-level filesystem error codes
- preserve `ENOTEMPTY`, `EISDIR`, and `ENOTDIR` through S0FS, ctld portal, and FUSE mappings
- keep generic `FailedPrecondition` errors mapped to `EIO`
- use operation-specific `unlink(2)` and `rmdir(2)` behavior for rootfs-backed portals
- replace the volume-file API error-message match with structured errno handling
- centralize the shared S0FS error mapping and add regression coverage

## Validation

- `make proto manifests apispec`
- `go mod tidy`
- `go vet ./storage-proxy/pkg/fserror ./pkg/volumefuse ./storage-proxy/pkg/fsserver ./storage-proxy/pkg/http ./ctld/internal/ctld/portal`
- targeted `go test -race -count=1` for all changed packages
- CI-equivalent non-e2e race suite completed with one unrelated `netd/pkg/proxy` SSH EOF; the failed test passed on an isolated race rerun
- remote kind mounted-volume verification will be posted while PR CI runs

### DIFF
--- a/ctld/internal/ctld/portal/rootfs_session.go
+++ b/ctld/internal/ctld/portal/rootfs_session.go
@@ -267,7 +267,7 @@ func (s *rootFSBackedSession) Unlink(_ context.Context, req *pb.UnlinkRequest) (
 	} else if preserved {
 		return &pb.Empty{}, nil
 	}
-	if err := os.Remove(s.hostPath(rel)); err != nil {
+	if err := unix.Unlink(s.hostPath(rel)); err != nil {
 		return nil, mapRootFSBackedError(err)
 	}
 	s.dropPath(rel)
@@ -283,7 +283,7 @@ func (s *rootFSBackedSession) Rmdir(_ context.Context, req *pb.RmdirRequest) (*p
 	if err != nil {
 		return nil, err
 	}
-	if err := os.Remove(s.hostPath(rel)); err != nil {
+	if err := unix.Rmdir(s.hostPath(rel)); err != nil {
 		return nil, mapRootFSBackedError(err)
 	}
 	s.dropPathTree(rel)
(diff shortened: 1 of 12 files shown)


laotoutou:
Remote validation completed against PR head `c35b26e0` on the persistent Aliyun Singapore kind environment.

- `Sandbox0Infra/fullmode`: `Ready 9/9`; default template pod: `1/1 Running`
- Created a temporary template, S0FS SandboxVolume, and sandbox; claim reported `bootstrap_mount_state=mounted`
- In the sandbox, through the real ctld/FUSE path:
  - S0FS: `rmdir(non-empty) -> errno 39 ENOTEMPTY`
  - S0FS: `unlink(directory) -> errno 21 EISDIR`
  - S0FS: `rmdir(file) -> errno 20 ENOTDIR`
  - rootfs returned the same three POSIX errno values
- Each rejected operation left its target intact
- Temporary sandbox, volume, and template were cleaned up; the persistent kind cluster was retained

PR #708, opened 2026-07-16
feat: consolidate storage-proxy into manager and netd into ctld

laotoutou:
## Summary

- run the storage-proxy API/runtime inside the manager process while preserving the storage-proxy Service, internal-auth audience, routes, and wire contracts
- run netd only in the elected ctld primary and bind netd failure to the existing ctld HA lease lifecycle
- migrate existing standalone workloads through guarded storage cutover and a node-local netd active lock with a bounded legacy fallback
- remove standalone operator reconciliation/RBAC after cutover, and update docs, samples, CI diagnostics, and E2E assertions
- persist canonical S0FS snapshot state in object storage so storage-proxy-to-manager Pod replacement cannot invalidate existing snapshots

## Compatibility and migration

- `pkg/apispec/openapi.yaml` is unchanged
- manager waits for a complete storage-enabled rollout and manager-only EndpointSlices before deleting the old storage-proxy Deployment
- existing storage-only specs remain valid; manager becomes the process host without implicitly enabling ctld or manager node-readiness gates
- differing manager/storage replica settings converge to the larger count; colliding legacy listen ports are remapped only inside the Pod while the compatibility Service port stays unchanged
- existing netd is first upgraded to a guarded current image, then rolled to a delayed/retryable fallback; it is deleted only after both ctld slot templates embed guarded netd, initial sync, CSI registration, and node readiness all converge
- S0FS snapshot objects are written before catalog insertion and read remote-first; local JSON is only a cache
- pre-change snapshots with lost local cache use a fail-closed recovery path bounded by a stable committed head, manifest invariants, segment existence, and a unique timestamp/size match before durable backfill
- E2E teardown quiesces only single-cluster template syncing through a process-local barrier, keeping manager-hosted storage and sandbox finalizers alive until namespace volume unpublish completes
- rollback to a pre-consolidation release is documented as a maintenance workflow to avoid overlapping an old unguarded netd with the embedded runtime

## Validation

- all Go package tests except launching E2E scenario packages
- targeted race tests across storage runtime/S0FS/snapshots, manager, ctld/netd, operator plan/reconcilers, framework, and E2E case helpers
- `go vet ./...`, `go build ./...`, and golangci-lint v2.5.0 (`0 issues`)
- `pkg/apispec/openapi.yaml` zero diff
- remote kind upgrade, legacy and new snapshot restore, storage/network compatibility, and ctld HA failover evidence attached in the PR discussion

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -64,6 +64,12 @@ jobs:
         run: |
           echo "=== pods ==="
           kubectl get pods -A -o wide || true
+          echo "=== consolidated workloads ==="
+          kubectl -n sandbox0-system get deployments,daemonsets,services -o wide || true
+          echo "=== manager/storage logs ==="
+          kubectl -n sandbox0-system logs -l app.kubernetes.io/component=manager --all-containers --tail=-1 || true
+          echo "=== ctld/netd logs ==="
+          kubectl -n sandbox0-system logs -l app.kubernetes.io/component=ctld --all-containers --tail=-1 || true
           echo "=== sandbox0-system events ==="
           kubectl -n sandbox0-system get events --sort-by=.lastTimestamp || true
           echo "=== infra-operator events ==="
--- a/.github/workflows/netd-cni-e2e.yml
+++ b/.github/workflows/netd-cni-e2e.yml
@@ -8,9 +8,12 @@ on:
       - "Dockerfile"
       - "Dockerfile.local"
       - "Makefile"
+      - "ctld/**"
+      - "infra-operator/**"
       - "manager/**"
       - "netd/**"
       - "pkg/**"
+      - "storage-proxy/**"
       - "tests/e2e/**"
       - "go.mod"
       - "go.sum"
@@ -79,8 +82,10 @@ jobs:
           kubectl get nodes -o wide || true
           echo "=== pods ==="
           kubectl get pods -A -o wide || true
-          echo "=== netd logs ==="
-          kubectl -n sandbox0-system logs -l app.kubernetes.io/name=netd --all-containers --tail=-1 || true
+          echo "=== consolidated workloads ==="
+          kubectl -n sandbox0-system get deployments,daemonsets,services -o wide || true
+          echo "=== ctld/netd logs ==="
+          kubectl -n sandbox0-system logs -l app.kubernetes.io/component=ctld --all-containers --tail=-1 || true
           echo "=== sandbox0-system events ==="
           kubectl -n sandbox0-system get events --sort-by=.lastTimestamp || true
           kind export logs "${RUNNER_TEMP}/kind-logs-${{ matrix.short }}" --name "${KIND_CLUSTER_NAME}" || true
--- a/README.md
+++ b/README.md
@@ -163,26 +163,24 @@ Sandboxing reduces blast radius and gives policy a real enforcement point. It do
 ```mermaid
 flowchart TB
     client["Client, SDK, CLI, or agent platform"] --> cgw["cluster-gateway"]
-    cgw --> mgr["manager"]
+    cgw --> mgr["manager<br/>lifecycle + storage-proxy runtime"]
     cgw --> pod["sandbox pod with procd"]
     mgr --> pod
-    mgr --> ctld["ctld (node-local)"]
-    mgr <--> netd["netd (optional)"]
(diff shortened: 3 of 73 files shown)


laotoutou:
## Remote upgrade validation (Aliyun persistent kind)

Validated PR head `25e2daf2` with image `sandbox0ai/infra:pr708-25e2daf2` by upgrading an existing old-topology cluster.

### Upgrade and final topology

- Manager converged first; ctld slot B reached 2/2 before slot A started, then slot A reached 2/2.
- `Sandbox0Infra/fullmode` is `Ready 9/9` and every sandbox0 core workload uses the PR image.
- No standalone netd DaemonSet and no standalone storage-proxy Deployment remain.
- Compatibility contracts remain routed:
  - `fullmode-storage-proxy:8081` selects the manager endpoint.
  - `fullmode-netd-metrics:9091` exposes both node IPs and returns Prometheus metrics.
- Both ctld DaemonSets use `maxUnavailable=1`, `maxSurge=0`; both data-plane nodes report `data-plane-ready=true`, `netd-ready=true`, and `ctld-ready=true`.

### Storage and snapshot continuity

- Existing volume, snapshot, and file APIs all returned 200 after the topology upgrade.
- Restoring the pre-upgrade legacy snapshot returned the original pre-upgrade file contents; strict legacy recovery selected immutable manifest sequence 2 and backfilled the durable snapshot object.
- Deleted/recreated the manager Pod, then restored the same old snapshot again: 200 with zero legacy-recovery warnings in the new manager Pod, proving the durable snapshot object survived cache/Pod replacement.
- Created a new snapshot (201), mutated the volume, recreated manager, restored the snapshot (200), then deleted it (200; subsequent GET 404).

### ctld/netd HA

- Deleted the active slot-B ctld on one node while running 120 sequential sandbox egress requests: 120/120 succeeded; slot A acquired the node-local netd lock and kept syncing redirect rules.
- Repeated in the reverse direction by deleting active slot A at request 3 of a strict 20-request probe: 20/20 succeeded; slot B acquired the lock at `2026-07-16T11:53:00Z`, the replacement A Pod became Ready, and node readiness stayed true.
- Metrics remained queryable after takeover.

Remote temporary probes and port-forwards were cleaned up.

laotoutou:
Latest fix b4934826 addresses the prior core E2E teardown timeout at its source. After storage-proxy moved into manager, cleanup had to keep manager running for CSI unpublish, but the live SingleClusterReconciler recreated deleted tpl-* namespaces from PostgreSQL desired state. Teardown now sends SIGUSR1 to every stable manager process, waits for a post-drain acknowledgement, deletes manager-owned namespaces while embedded storage and sandbox finalizers remain available, and only then scales manager down. The barrier blocks new template work and drains in-flight reconciliation without changing OpenAPI. Targeted race tests, full non-E2E tests, vet, build, lint, and diff checks pass; the new CI run is in progress.

laotoutou:
Remote follow-up on latest head b4934826 (Aliyun Singapore persistent kind, two workers): built and loaded sandbox0ai/infra:pr708-b4934826 plus its procd-bin image, upgraded Helm to revision 7, and reached Sandbox0Infra Ready 9/9 with manager, cluster-gateway, ctld-a, and ctld-b all on the latest tag.

Teardown barrier validation reproduced the prior core-CI boundary directly. The manager advertised the support marker with no stale ack; storage-proxy /healthz and manager /readyz were healthy. After SIGUSR1, the manager emitted the quiesce ack and both Reconciler quiesced / Template reconciliation quiesced logs. I deleted tpl-default and checked it remained absent for 45 seconds, beyond the configured 30-second reconciliation period. During that window the same manager Pod UID stayed Ready with restartCount 0, and both embedded storage /readyz and manager /readyz remained healthy.

I then restarted manager. A new Pod UID came up on b4934826 with the support marker and no quiesced marker, tpl-default and its 1/1 default pool Pod were recreated, storage /readyz remained healthy, and Sandbox0Infra returned to Ready 9/9. Final topology has zero standalone storage-proxy Deployments and zero standalone netd DaemonSets; the compatibility storage Service targets manager endpoints, netd metrics targets ctld slot A endpoints on both nodes, and metrics are readable.

laotoutou:
Final CI status for b4934826: quick-check, core E2E, s0fs-posix, mountpoint-s3, all four netd CNI matrices (kindnet, flannel, canal, cilium native-host legacy), release metadata, release infra, and the aggregate e2e check all passed. The publish-images job is the expected PR skip. PR is mergeable; review approval is the only remaining gate.

PR #709, opened 2026-07-16
refactor: introduce canonical fullmode runtime config

laotoutou:
## Summary

- add spec.storage.runtime and spec.network as canonical data-plane runtime configuration
- keep services.storageProxy and services.netd as temporary fallback aliases for the staged rollout
- route canonical storage traffic through the manager Service while preserving the legacy alias path
- migrate fullmode and multi-cluster data-plane samples to the canonical fields
- preserve combined manager/storage resource defaults and explicit total process budgets

The public OpenAPI contract is unchanged.

## Validation

- make manifests
- go test -count=1 ./infra-operator/...
- helm lint infra-operator/chart
- parsed updated YAML samples
- git diff --check
- verified pkg/apispec/openapi.yaml SHA-256 is unchanged


laotoutou:
Remote kind validation on PR head `392f1839` passed:

- `Sandbox0Infra` Ready with all 9 transition conditions current
- manager `storage-http:8081` EndpointSlice ready
- ctld A/B both 2/2; no standalone storage-proxy Deployment or netd DaemonSet
- login, team query, sandbox claim/readiness, synchronous command execution
- sandbox file read/write
- volume file read/write and snapshot create/overwrite/restore/list/delete
- network allow-all, block-all applied-hash convergence, allowed `example.com`, blocked Google, then reset to allow-all
- sandbox and volume cleanup

The persistent remote ECS has been stopped in StopCharging mode after validation.

PR #713, opened 2026-07-16
refactor: finalize fullmode runtime architecture

laotoutou:
## Summary

- make spec.storage.runtime and spec.network the only data-plane runtime configuration
- run the storage API in manager and the network engine in the ctld A/B process
- remove standalone storage-proxy/netd binaries, reconcilers, RBAC, workloads, services, legacy deployment samples, and degraded-mode E2E cases
- keep only fullmode deployment samples and rewrite architecture/operations documentation around the steady-state design
- use the ctld HA lease as the single network-runtime fence and expose network metrics through the ctld service identity

## Safety and rollout

- public OpenAPI source is byte-for-byte unchanged (SHA-256: 6a895b8f208e09f5e428ed5b659cdbb60d37e8596d0998001870f2fb575ab485)
- the manager storage auth transition was deployed and production-smoked first
- this PR retains only the narrow rolling auth/resource cleanup hooks needed for one production rollout; the final follow-up removes them after this image is live
- manager service port collisions and stale ctld rollout revisions are rejected explicitly

## Verification

- make manifests (idempotent)
- git diff --check
- broad non-E2E Go test sweep plus targeted operator/controller/plan/auth/cluster-gateway/manager/ctld tests
- CGO_ENABLED=0 manager and ctld builds
- Helm lint and all sample YAML parsing
- local E2E intentionally not run; GitHub CI and the remote kind environment own E2E validation


laotoutou:
Remote kind validation passed for the current runtime diff.

- fullmode reached Ready 10/10
- no standalone storage-proxy/netd workloads, Services, or RBAC remained
- manager storage HTTP/ready endpoints passed on port 8081
- ctld A/B was Ready on both sandbox workers; ctld-network-metrics selected the active slot
- volume direct I/O, snapshot restore, sandbox mount read/write, and persisted data passed
- network exact-allow, block-all, and allow-all recovery passed with desired/applied hashes matching
- test resources returned 404 after cleanup; component restart counts stayed at zero
- persistent kind/cache was retained and the ECS host was stopped with StopCharging

The final head d77a3dcc only removes unused test helpers after the runtime image build; the remote source tree was resynced and verified clean at d77a3dcc. No local E2E was run; GitHub CI owns the E2E matrices.

laotoutou:
Follow-up: current head a08b3c4a differs from the remotely validated runtime only by correcting an existing ordered lifecycle E2E assertion: after the Retain transition, the Delete transition now checks the retained Secret/PVC instead of requiring the already-removed StatefulSet/Service. Runtime code and the remote validation result are unchanged. The package compiles and golangci-lint v2.5.0 reports 0 issues.

PR #715, opened 2026-07-16
refactor: remove consolidated runtime transition compatibility

laotoutou:
## Summary

- remove the one-release storage-proxy audience/caller compatibility and superseded operator condition, node-label, and metrics-Service cleanup paths
- make manager storage and ctld network runtime the only current process model in code, generated CRDs, tests, logs, and workflow display text
- retain stable public/wire/storage/metric identities, including the byte-identical public OpenAPI contract

## Validation

- CI-equivalent non-E2E Go suite with `-race -count=1`
- golangci-lint v2.5.0
- generated artifacts idempotent; `go mod tidy` clean
- Helm lint and changed YAML parse
- OpenAPI SHA-256 unchanged: `6a895b8f208e09f5e428ed5b659cdbb60d37e8596d0998001870f2fb575ab485`

Remote kind validation will be posted separately after CI starts.


laotoutou:
## Remote kind validation (Aliyun Singapore)

Validated exact head `a59446b42fabffc5fad760462ca1d1b74073cde6` on the persistent `sandbox0-e2e` kind environment.

- Fullmode reached `Ready 10/10`; `StorageRuntimeReady=True` and `NetworkReady=True`.
- Manager owns the storage endpoint: service/EndpointSlice port `storage-http:8081` is ready and `/healthz` returned `healthy`.
- ctld topology is A/B on both sandbox workers (4 Pods total), exactly one container per Pod.
- No standalone `storage-proxy` or `netd` workload, Service, service account, Role/Binding, or ClusterRole/Binding remains.
- Manager, cluster-gateway, and all ctld Pods ran image ID `sha256:054f02f15982477d7527813be38297fa279ddc9f9397f423c70502d2f2c687d4`.

Functional smoke passed:

- direct SandboxVolume file write/read;
- snapshot after mutation restored the original bytes;
- claimed sandbox mounted the volume and passed host/API -> sandbox read plus sandbox -> direct API read;
- network initial allow, block-all deny, explicit CIDR/port allow, and allow-all recovery all passed, with desired/applied hashes matching.

HA fault injection passed:

- on `sandbox0-e2e-worker`, slot A (`fullmode-ctld-a-bg8pk`) was the active ctld and slot B (`fullmode-ctld-b-d9njh`) was standby;
- deleting active A promoted the existing B in the same recorded second (Killing and “primary started network runtime” at `22:02:18Z`, observed within the 1-second poll interval);
- block-all remained effective through promotion;
- the existing FUSE mount retained its pre-failover file and accepted a post-failover write that was readable through the direct volume API;
- a post-promotion network policy update applied and restored connectivity;
- DaemonSet A recreated as `fullmode-ctld-a-p5xtj`; A/B returned Ready with one container each.

Cleanup/audit:

- test sandboxes: `[]`; test volumes: `[]`; temporary namespaces: none;
- current platform Pods are Running with `restartCount=0` (the active ctld deletion above was intentional);
- no panic/fatal-level log entries were found after recovery.

Environment note: the first storage attempt was correctly rejected by the unchanged 5 GiB s0fs free-space guard because the persistent test host had ~3.97 GiB projected free. I removed only stale, unused host Docker/gVisor cache, restored ~7.4 GiB available, deleted the orphaned test volume, and ran the successful matrix from a clean state.

PR #721, opened 2026-07-17
fix(ctld): report volume storage metering

laotoutou:
## Summary

- route S0FS sync and compaction through a shared VolumeContext materialization API that always invokes storage observation
- share canonical volume observation construction and outbox watermark recording between manager and ctld
- wire the regional PostgreSQL metering outbox into ctld volume portals when metering is enabled
- make the operator build manager and ctld storage runtime configuration through the same storageruntime.BuildRuntimeConfig path
- consolidate ctld S0FS VolumeContext construction so bind and owner-only mounts cannot diverge again

## Root cause

The embedded manager storage runtime observed every materialized manifest, but ctld portal paths called the S0FS engine directly. Writes served through ctld were durable in S3 but did not update the volume storage projection used by metering.

A second duplicated path existed in the operator: ctld independently assembled StorageProxyConfig and omitted metering, while manager used the shared storage-runtime builder. As a result, ctld still received metering.enabled=false even when the Sandbox0Infra enabled metering.

## Validation

- go test ./infra-operator/internal/controller/services/...
- go test ./storage-proxy/pkg/... ./ctld/internal/... ./ctld/cmd/...
- go vet ./infra-operator/internal/controller/services/ctld ./infra-operator/internal/controller/services/storageruntime ./storage-proxy/pkg/volume ./storage-proxy/pkg/snapshot ./ctld/internal/ctld/portal ./ctld/cmd/ctld
- remote persistent kind environment with builtin ClickHouse and metering enabled:
  - generated ctld ConfigMap had metering.enabled=true
  - created and mounted a SandboxVolume through ctld
  - wrote and read back 131072 bytes through the sandbox file API
  - mount ownership was ctld-node/sandbox0-e2e-worker
  - PostgreSQL storage projection reported 131072 bytes and emitted two outbox rows
  - ClickHouse storage_projection_state contained the 131072-byte volume observation

### DIFF
--- a/ctld/cmd/ctld/main.go
+++ b/ctld/cmd/ctld/main.go
@@ -26,12 +26,15 @@ import (
 	"github.com/sandbox0-ai/sandbox0/pkg/ctldapi"
 	"github.com/sandbox0-ai/sandbox0/pkg/dbpool"
 	"github.com/sandbox0-ai/sandbox0/pkg/k8s"
+	meteringoutbox "github.com/sandbox0-ai/sandbox0/pkg/metering/outbox"
+	"github.com/sandbox0-ai/sandbox0/pkg/naming"
 	"github.com/sandbox0-ai/sandbox0/pkg/observability"
 	httpobs "github.com/sandbox0-ai/sandbox0/pkg/observability/http"
 	"github.com/sandbox0-ai/sandbox0/pkg/sandboxprobe"
 	"github.com/sandbox0-ai/sandbox0/pkg/volumeportal"
 	storagedb "github.com/sandbox0-ai/sandbox0/storage-proxy/pkg/db"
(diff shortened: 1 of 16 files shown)


laotoutou:
Final validation on c67669a0:

- all PR checks passed, including quick-check, core E2E, mountpoint-s3 E2E, s0fs-posix E2E, and all four network CNI jobs
- persistent remote kind deployment produced a ctld ConfigMap with metering.enabled=true
- a 131072-byte file written and read through the sandbox file API/FUSE path produced a 131072-byte PostgreSQL storage projection, two outbox rows, and a matching ClickHouse storage_projection_state row
- the active mount owner was ctld-node/sandbox0-e2e-worker

Remote test resources were cleaned after validation; the persistent kind cluster was preserved.

PR #759, opened 2026-07-26
perf(manager): streamline hot claim reservations

laotoutou:
## What

- reduce the foreground hot-claim Pod mutations from two PATCHes to one by completing initialization through the durable sandbox record and leaving warm-pool detachment to the existing background controller
- reserve informer-cached idle candidates inside each manager process until the informer observes the Kubernetes reservation, preventing concurrent claims from repeatedly selecting the same stale candidate
- give latency-sensitive hot-claim PATCH/GET/resize calls an independent Kubernetes client token bucket so cleanup, refill, and detachment traffic cannot consume their client-side budget
- remove `resourceVersion` from the background detachment CAS while retaining UID, pool label, reservation token, and reservation state preconditions

## Why

At burst concurrency, claims could collide on one stale informer candidate, contend on a shared client-side Kubernetes rate limiter, and perform a second foreground Pod PATCH after procd initialization. These costs showed up directly in burst TTI and amplified API-server contention.

The durable sandbox record now moves from `starting` to `running` only after persistence and procd initialization complete. A versioned Pod annotation tells the new reservation controller that this record transition is the completion signal. Legacy `ready` reservations remain supported, and an initializing legacy reservation is not inferred complete from a running record.

## Safety

- Kubernetes metadata CAS remains the cross-process ownership boundary; the local allocator only removes same-process duplicate selection.
- The completion protocol marker prevents mixed-version forward rollouts from detaching legacy initializing claims prematurely.
- A crash before the durable `running` transition remains recoverable as an abandoned reservation; a crash after it lets the controller safely detach the completed claim.
- Background detachment still verifies Pod UID, idle-pool label, reservation token, exact state, and the matching durable sandbox identity/generation.

## Validation

Local:

- `CONFIG_PATH=/dev/null go test ./manager/pkg/service ./manager/cmd/manager -count=1`
- `CONFIG_PATH=/dev/null go test -race ./manager/pkg/service -count=1`
- `go vet ./manager/pkg/service ./manager/cmd/manager`
- `CONFIG_PATH=/dev/null go test ./manager/... -count=1`

Persistent Aliyun Singapore kind environment:

- fullmode deployment became Ready; manager used the branch image with the dedicated hot-claim client enabled
- one 128 MiB default claim passed command execution, binary file write/read, deletion, and idle-pool replenishment
- a persistent 128 MiB test template with 16 ready idle Pods handled 16 simultaneous hot claims: 16/16 HTTP 201 in 908 ms wall time, 16 unique sandbox IDs, and the claimed Pod set exactly matched the 16 prepared idle Pods
- all 16 durable records reached `running`; all 16 reservations detached; no reservation/protocol annotations remained
- metrics reported 16 `reserved`, 16 `success`, no `update_conflict`, and 16 hot completions; manager remained Ready with zero restarts and no client-side 429/throttle or panic
- all temporary sandboxes/template resources were deleted and the remote ECS instance was returned to `Stopped/StopCharging`

## ComputeSDK benchmark

Private ComputeSDK commit `15b1a338171f5631c266bb480e8491cba3f6a4e3`, 100 iterations per mode:

| Mode | Score | Success | Median | P95 | P99 | Wall | First ready |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| sequential | 97.01 | 100/100 | 214.95 ms | 402.91 ms | 464.64 ms | — | — |
| staggered | 96.03 | 100/100 | 294.49 ms | 541.22 ms | 567.71 ms | 20.89 s | 185.06 ms |
| burst 1 | 79.74 | 100/100 | 1,854.21 ms | 2,262.82 ms | 2,321.11 ms | 4.89 s | 965.63 ms |
| burst 2 | 82.09 | 100/100 | 1,385.17 ms | 2,389.65 ms | 2,414.42 ms | 3.80 s | 742.99 ms |

Compared with the previous same-boundary PR #758 run:

- staggered: score +5.74, median -67.88%, P95 -48.02%
- burst 1: score +10.24, P95 -49.32%, wall -27.79%
- burst 2: score +20.25, median -42.75%, P95 -57.90%, wall -50.55%

Both burst rounds are retained. The result measures the three changes together and does not isolate the contribution of one change.

Exact boundary:

- image `sandbox0ai/infra:pr759-a146226`, digest `sha256:7fdba52d958c615c05f8dc4c9d8a4aef94e6bf02a4473d596de3fd57dce79b7f`
- [image CI](https://github.com/sandbox0-ai/sandbox0/actions/runs/30218755018)
- [successful Ali us-east-1 GitOps deployment](https://github.com/sandbox0-ai/sandbox0-infra/actions/runs/30219687731)
- one original fixed `ecs.r9ae.2xlarge` sandbox node, 8 vCPU / about 64 GiB
- default pool `minIdle=maxIdle=120`; all 120 idle Pods Ready at `128Mi/150m`
- benchmark sandboxes explicitly requested `memory=128Mi`, avoiding resize
- burst capacity locked to `0-0`; actual burst nodes remained zero for all four formal runs
- the temporary runner called the in-cluster regional gateway, removing external network effects
- no ComputeSDK ingest address or key was set and no upload workflow was called

The deployed manager metrics exactly covered the runner plus the four formal rounds: 401/401 successful default hot claims, 401 idle reservations, 401 idle successes, and no `update_conflict` or error series. Mean manager claim time was 262.04 ms (`claim_idle_pod` 163.84 ms; reservation completion 21.96 ms). Shared and hot-claim Kubernetes clients both reported their independent `50 QPS / 100 burst` budgets.

The runner was deleted, the three temporary team quota overrides were removed, and burst capacity was restored to `0-299` with zero actual burst nodes. Final preflight: one original fixed node, idle/Ready `120/120`, active 0, pending 0, idle restart 0; exact PR image remained Ready across all core workloads.

After burst 1, two newly refilled idle Pods each self-recovered from one gVisor `StartError` caused by an incomplete mounts JSON read. They were removed and replaced before burst 2 so the baseline returned to idle restart 0. This is a separate ctld/rootfs atomic-write issue, not a manager hot-claim path change in this PR.

### DIFF
--- a/manager/cmd/manager/main.go
+++ b/manager/cmd/manager/main.go
@@ -124,6 +124,16 @@ func main() {
 	if err != nil {
 		logger.Fatal("Failed to create Kubernetes client", zap.Error(err))
 	}
+	hotClaimK8sConfig := isolatedK8sClientConfig(k8sConfig)
+	observeHotClaimK8sClientRateLimit(managerMetrics, hotClaimK8sConfig)
+	logger.Info("Hot claim Kubernetes client rate limit configured",
+		zap.Float32("qps", effectiveK8sClientQPS(hotClaimK8sConfig)),
+		zap.Int("burst", effectiveK8sClientBurst(hotClaimK8sConfig)),
+	)
+	hotClaimK8sClient, err := kubernetes.NewForConfig(hotClaimK8sConfig)
+	if err != nil {
+		logger.Fatal("Failed to create hot claim Kubernetes client", zap.Error(err))
+	}
 
 	// Add SandboxTemplate to scheme
 	if err := v1alpha1.AddToScheme(scheme.Scheme); err != nil {
@@ -461,6 +471,7 @@ func main() {
 	sandboxService.SetCredentialStore(credentialStore)
 	sandboxService.SetQuotaStore(quotaRepo)
 	sandboxService.SetSandboxStore(sandboxStore)
+	sandboxService.SetHotClaimK8sClient(hotClaimK8sClient)
 	hotClaimReservationController := service.NewHotClaimReservationController(
 		k8sClient,
 		podLister,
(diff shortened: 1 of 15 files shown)


laotoutou:
CI note: `quick-check` failed twice in the existing `cluster-gateway/pkg/http` audit-delivery timing tests, outside this PR's manager-only change surface.

- attempt 1: `TestAuditDeliveryPrioritizesForegroundOverReplay` (`canonical batch sizes = [1]`)
- attempt 2: `TestAuditDeliveryRecordsStageAndBatchMetrics` (`canonical in-flight = 1, want 0`)
- `cluster-gateway/pkg/http/audit.go` and `audit_delivery_test.go` are identical to `origin/main`
- the attempt-1 test reproduced 4/100 under `go test -race` on the unchanged main code
- the attempt-2 test passed 100/100 in isolation, and only failed under the full repository race run, which is consistent with the same timing-sensitive suite flake

All manager packages passed in both CI attempts. Core e2e, s0fs, mountpoint-s3, every network matrix job, and release metadata/infra checks are green. I am intentionally not mixing an unrelated audit test fix into this performance PR.

PR #766, opened 2026-07-28
Batch metering projections and guard ClickHouse OSS cleanup

laotoutou:
## Summary

- coalesce consecutive PostgreSQL metering outbox transactions into bounded batches of at most 500 operations
- issue multi-row ClickHouse inserts per projection table, preserve storage mutation order, and advance watermarks only after all data/state writes succeed
- keep PostgreSQL as the synchronous producer boundary; audit delivery code and its canonical-sync acknowledgement contract are unchanged
- add a concurrent pending-batch index so batching does not block producer writes
- add a guarded, exact-key OSS reconciler with separate scan/apply phases, live-path rechecks, age/fraction bounds, and 1,000-key delete batches

## Production evidence

The one-time guarded reconciliation for `clickhouse/ali-ue1/` removed 147,740 confirmed orphan objects (4,466,232,139 bytes). A fresh two-snapshot scan found zero remaining candidates older than 72 hours, while recent metering and audit rows remained queryable. This PR does not deploy or restart any service.

## Verification

- `go test ./pkg/metering/... ./internal/clickhouseobjectgc ./tools/clickhouse-oss-gc`
- PostgreSQL 16 integration tests for migrations, coalesced claims, strict head ordering, retries, and atomic acknowledgement
- `go vet ./pkg/metering/... ./internal/clickhouseobjectgc ./tools/clickhouse-oss-gc`
- full `go test -p 2 ./...` package/unit run passed until the Kind e2e suite, which could not start because `kind` is not installed in this workspace

### DIFF
--- a/internal/clickhouseobjectgc/reconciler.go
+++ b/internal/clickhouseobjectgc/reconciler.go
@@ -1,0 +1,288 @@
+package clickhouseobjectgc
+
+import (
+	"context"
+	"fmt"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/s3"
+	"github.com/aws/aws-sdk-go-v2/service/s3/types"
+)
+
+const (
+	ManifestVersion    = 1
+	maxDeleteBatchSize = 1000
+)
+
+type objectClient interface {
+	ListObjectsV2(context.Context, *s3.ListObjectsV2Input, ...func(*s3.Options)) (*s3.ListObjectsV2Output, error)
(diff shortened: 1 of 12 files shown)


laotoutou:
Additional compatibility check: ClickHouse 24.8.14.39 accepted the generated multi-row `INSERT ... SETTINGS async_insert = 0, wait_for_async_insert = 1 VALUES ...` syntax and returned the expected rows.

laotoutou:
Final CI status: all required checks are green. The mountpoint-s3 E2E passed on attempt 3 (job 90289434635, 16m28s), including all compatibility assertions and cleanup. The two prior retries had reached the existing namespace-cleanup race after functional assertions completed; no product-code changes were made for that unrelated flake.

PR #800, opened 2026-08-02
Add remote-resize VNC server to coding-agent image

laotoutou:
## Summary

- install TigerVNC in the coding-agent template so browser takeover can negotiate the remote desktop size
- keep Xvfb/x11vnc temporarily for rolling compatibility with the currently deployed Sandpi runtime
- assert the image contains `Xtigervnc` during both tests and image build

## Validation

- `go test ./manager/procd`
- all non-e2e/non-integration Go packages
- `go vet ./...`
- `git diff --check`

Image-level and browser takeover validation will run after publishing the template image and on the persistent remote test environment.

### DIFF
--- a/manager/procd/Dockerfile.coding-agent
+++ b/manager/procd/Dockerfile.coding-agent
@@ -21,7 +21,8 @@ ENV DEBIAN_FRONTEND=noninteractive \
 # Claude Code and Pi require Node.js 22. Pi currently requires Node.js 22.19+
 # specifically, so fail the build if the repository ever resolves an older 22.x.
 RUN apt-get update \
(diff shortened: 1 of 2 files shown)


laotoutou:
CI note: PR Quick Check passed, as did the s0fs-posix and mountpoint-s3 E2E suites. Core E2E failed twice on the same unrelated sandbox webhook assertion: 46/47 specs passed, while `sandbox.deleted` was not observed within 90 seconds (`tests/e2e/cases/api_mode_suite.go:655`). This PR only changes `manager/procd/Dockerfile.coding-agent` and a source-contract unit test; it does not change manager, sandbox lifecycle, or webhook code. The same base commit passed core E2E earlier today in PR #799. Proceeding with an admin merge and validating the actual image/runtime path after publication.

PR #821, opened 2026-08-03
revert(rootfs): restore legacy persistent rootfs

laotoutou:
## Summary

- fully revert #815 and remove the OCI marker/rootfs snapshotter architecture
- restore the legacy `gvisor-rootfs` pause/resume implementation and operator contract
- remove the associated runtime, schema, documentation, and generated artifacts

## Validation

- the resulting source tree is byte-for-byte identical to `a09af4402bd42d4f1928fa21ea18adef36dd4bfe`
- production rollback smoke test: 203/203 warm pods Ready on `gvisor-rootfs`
- coding-agent create: 1.365s
- full pause: 1.233s
- resume to `running`: 1.814s; resume to successful exec: 2.633s
- persisted file content survived pause/resume

## Deployment

- production runtime is already pinned to `sandbox0ai/infra:main-a09af44`
- infra-operator and GitOps manifests are pinned to `a09af4402bd42d4f1928fa21ea18adef36dd4bfe`


laotoutou:
Revert rationale:

#815 met its large-rootfs hot-path performance goal in the tested cases:

- 512 MiB incompressible: 284 ms durable pause, 1.909 s resume API, 2.714 s Ready
- 2 GiB incompressible: 286 ms durable pause, 2.103 s resume API, 2.927 s Ready

The rollback is based on operational and architectural risk discovered during the production rollout, not a failure of those latency benchmarks:

- it introduced a second runtime and persistence model (`sandbox0-rootfs` alongside `gvisor-rootfs`), with an explicit upgrade boundary for existing paused data;
- safe rollout and rollback required coordinated state across manager, ctld, infra-operator, CRDs, containerd configuration, the external snapshotter, RuntimeClasses, GitOps, and node-pool images;
- mixed-version nodes and controllers could produce incompatible warm-pool Pods or continuously overwrite manual recovery state;
- the clean test environment did not represent production's existing persisted data, already-running sandboxes, and nodes created from different runtime templates;
- the resulting migration and operating model was judged too complex to keep, even though the benchmarked hot path itself was fast.

We are restoring the known `gvisor-rootfs` runtime/data path, including its existing capacity-linear limitation, rather than operating two incompatible paths. A future redesign should retain #815's capacity-independent performance properties while providing a simpler single-path upgrade/rollback model, ideally by evolving `gvisor-rootfs` instead of introducing a second runtime family. It should continue to verify no claim-path registry dependency, S3 write amplification, and deletion reclamation.

PR #829, opened 2026-08-05
feat(rootfs): implement capacity-independent COW persistence

laotoutou:
## Summary

- replace the capacity-linear legacy rootfs tar/layer path with a v3 immutable metadata Head format
- keep sandbox writes on the native node-local overlay upper while exposing persisted data through a lazy read-only FUSE lower
- add continuous upper capture, bounded idempotent seal/acknowledge APIs, team-scoped immutable CAS, sparse files, hardlinks, xattrs, whiteouts, opaque directories, and deterministic metadata encoding
- add the external `sandbox0` containerd snapshotter behind the existing `gvisor-rootfs` runtime family, including marker materialization, node-local verified object cache, readiness, and process-instance recovery
- publish filesystem Heads through manager PostgreSQL transactions with CAS, capture/write leases, temporary prefix guards, asynchronous exact inventory, orphan recovery, durable deletion queues, snapshots, forks, restores, and storage metering
- replace legacy template capture with deterministic Head-to-OCI export and wire the snapshotter/runtime through infra-operator, generated CRDs/RBAC, kind configuration, and self-hosted documentation

## Why

The legacy implementation serializes and replays rootfs data in the pause/resume path, so lifecycle latency grows with persisted bytes and layer depth. This implementation continuously persists stable changes and makes a published Head represent the complete logical overlay state. Stable pause publishes only a bounded Head reference and the remaining dirty tail; resume attaches that Head without walking historical lineage.

This is a new implementation inspired by the capacity-independent direction evaluated in #815. It does not restore that PR's format or parallel runtime family.

## Architecture and consistency

- PostgreSQL `filesystem -> Head` is the publication truth; S3 contains immutable objects written before publication.
- Manager advances the Head, lifecycle transaction, and sandbox state in one PostgreSQL transaction using expected-parent CAS.
- Seal is idempotent and does not rely on ctld-local prepared JSON state.
- Team-scoped object keys allow same-team snapshot/fork sharing while preserving cross-team isolation.
- Exact object inventory and Head-to-OCI export are asynchronous, capacity-linear background work and are not part of the lifecycle publication hot path.
- Snapshotter process drift triggers a durable seal and runtime rebuild; it is intentionally not described as transparent FUSE reconnect.

## Breaking change and rollout

This is a one-way v3 format boundary. Migrations 13-16 add v3 state and remove the legacy online rootfs schema/path. Manager and ctld do not keep a legacy/v2 reader. Existing paused sandboxes and snapshots must be deleted or migrated offline during a maintenance window before cutover.

The rollout must install and validate the snapshotter/containerd handler on drained nodes before switching ctld and manager. Planned snapshotter replacement also requires node drain because it owns live FUSE mounts. A code rollback after new v3 data is published is not a data rollback.

## Validation

Local validation:

- `make test-all`
- `go vet ./...`
- rootfs format/capture/reader/FUSE/snapshotter/maintenance/export/sandboxstore/service race tests
- PostgreSQL migrations 1-16 and the complete rootfs v3/sandboxstore integration group
- Linux/arm64 static builds for ctld, manager, and rootfs-snapshotter
- Helm lint/template, generated manifests/API idempotency, and `git diff --check`
- temporary merge-tree against current `main`, followed by infra-operator, templateimage, and manager service tests

Remote Singapore kind validation covered:

- 512 MiB random file integrity, a 1 GiB logical sparse file, and 10,000 small files
- hardlinks, symlinks, binary xattrs, whiteouts, opaque directories, rename/unlink, socket masking, and excluded runtime paths
- pause/resume, snapshot/fork isolation, snapshot restore, snapshotter process drift recovery, deletion, exact inventory, and GC
- registry verification showing no persistent marker image pulls

Measured remote phases were approximately 20 ms pause API acceptance, 95 ms durable PostgreSQL rootfs lifecycle, 1.2 s old Pod deletion, and 5.75 s resume to product Ready. Exact inventory completed asynchronously in about 71 s for the 10,000-file workload and did not block pause.

## Known I/O boundaries

- the initial 1 MiB chunk size can amplify a cold 4 KiB lower read to 1 MiB; verified disk-cache hits avoid that object-store read
- changing any bytes in an upper regular file currently rereads and rechunks the complete file, although unchanged chunks deduplicate in CAS
- arbitrary directory seek replay is linear in the requested offset but bounded in memory; forward `readdir` streams shards
- FIFO metadata is representable, but gVisor guest-created FIFOs are not currently visible in the host overlay upper and therefore are not guaranteed to persist

These costs are documented and reported separately from stable lifecycle latency.


laotoutou:
Validation update for df13185c (merged-overlay event capture fix):

- Confirmed both ctld A/B receive /host-run/containerd with HostToContainer propagation, and the real task root mountinfo exposes the expected overlay upperdir.
- 10 immediate write -> pause -> resume cycles preserved /root/.bashrc exactly; durable pause completion was 135-154 ms (142 ms average).
- A 64 MiB random file written immediately before pause restored with the exact size and SHA-256; pause completed in 292 ms.
- 10,000 small files drained the capture queue in 22 ms; pause completed in 172 ms; all files restored with an identical tree hash.
- Template-upgrade investigation: a sandbox paused on default-v0.2.0 resumed after the default template moved to a compatible v2 image. Its saved .bashrc, 64 MiB file, and 10,000-file tree were preserved. The resumed Pod used the immutable rootfs-heads digest and did not inherit the v2-only file. A newly claimed control sandbox did use the v2 image and observed both v2 markers. This documents current behavior only: paused sandboxes remain pinned to their saved rootfs/base.
- All PR checks pass. The first kindnet runner became unresponsive in environment preparation; after force-cancel, the targeted PR rerun passed. A separate workflow_dispatch run on the same SHA also passed all four CNI matrices.

Remote test sandboxes were deleted, the default template was restored to default-v0.2.0, and the Singapore ECS was stopped in StopCharging mode.

PR #836, opened 2026-08-06
Derive builtin template CPU from configured ratio

laotoutou:
## What changed

- derive CPU for legacy builtin template presets from the configured memory-per-CPU policy before validation
- preserve explicit full builtin specs as validation-only contracts
- cover the 2Gi/CPU policy and explicit-spec behavior with focused tests

## Why

`teamTemplateMemoryPerCpu` is configurable, but builtin presets were fixed at 4Gi/CPU. Setting production to 2Gi/CPU therefore caused builtin synchronization to fail validation.

## Validation

- `go test ./infra-operator/internal/controller/... ./pkg/template/...`
- `git diff --check`

## Deployment dependency

This runtime change must be available in the production infra image before sandbox0-infra renders `teamTemplateMemoryPerCpu: 2Gi`.


laotoutou:
Reopening to retrigger CI after the GitHub Actions incident dropped the original pull_request events.

PR #851, opened 2026-08-14
fix(imagefs): skip superseded template revisions

laotoutou:
## Summary

- verify a claimed ImageFS revision still matches the current canonical template before creating its importer Pod
- terminate stale claims as `Superseded` instead of importing the current spec under an obsolete revision identity
- revive only superseded revisions when a template intentionally returns to an earlier spec

## Why

Production Stage 1 shadow import exposed two immutable revisions for `default`. The region-wide worker leased the older revision, but importer construction loads the current template spec. Without this guard, a stale revision can import a newer spec while retaining the stale revision ID and spec hash.

## Validation

- `go test ./manager/pkg/templateimagefs ./pkg/template ./pkg/template/store/pg`
- `go test ./manager/pkg/... ./pkg/template/...`
- `git diff --check`

## Rollout

The production green cluster remains shadow-only: S0FS admission is off, shared carrier pool is disabled, no revision is selected, and the latest status-only audit observed zero user sandbox and carrier Pods. After CI, this PR will be squash-merged, built without moving `latest`, then deployed only to green through the existing guarded GitOps workflow.

### DIFF
--- a/manager/pkg/templateimagefs/worker.go
+++ b/manager/pkg/templateimagefs/worker.go
@@ -34,6 +34,8 @@ const (
 	claimErrorLogPeriod = 30 * time.Second
 )
 
+var errTemplateImageRevisionSuperseded = errors.New("template ImageFS revision is superseded")
+
 type queue interface {
 	templstore.TemplateStore
 	templstore.TemplateImageRevisionStore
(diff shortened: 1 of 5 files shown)


laotoutou:
Additional validation: started an isolated local PostgreSQL 15 instance, ran `TEST_DATABASE_URL=... go test ./pkg/template/store/pg -run TestEnsureTemplateImageRevision -count=1 -v`, and both the shadow-selection and superseded-revival integration tests passed. The temporary database was stopped and moved to trash afterward.

PR #868, opened 2026-08-15
Remove Sandbox Volume and s0fs

laotoutou:
## What changed

- remove the public Sandbox Volume API, claim mounts, SDK contract, quota dimensions, metering producers, and documentation
- remove storage-proxy, s0fs, ctld FUSE/CSI portal code, operator configuration, benchmarks, and volume-specific tests
- keep rootfs persistence, snapshots, restore, forks, and encrypted object storage as the durable sandbox storage path
- move the shared object-store implementation into pkg/objectstore
- preserve storage_proxy migrations 00001-00016 byte-for-byte and add terminal migration 00017; add manager 00019, quota 00004, and metering 00008 without reusing rolled-back sequence numbers

## Migration behavior

The terminal migrations delete legacy Volume metadata and mutable projection state. Metering 00008 emits transactional ClickHouse tombstones before deleting PostgreSQL Volume/Snapshot state, so the query projection cannot remain active. Down sections intentionally do not recreate deleted capability or data. Historical storage migrations remain embedded so databases at any previously deployed version can reach the terminal migrations safely.

Manager migration 00019 intentionally skips the previously deployed and rolled-back 00013-00018 sequence.

## Validation

- full non-cluster Go suite: `go list ./... | rg -v /tests/e2e/scenarios/ | xargs go test`
- PostgreSQL integration tests for storage_proxy 00017, manager 00019 (including rolled-back 00013-00018 history), quota 00004, and metering 00008 ClickHouse tombstones
- `make apispec`
- `make manifests`
- `git diff --check`
- persistent Aliyun Singapore three-node Kind: Sandbox0Infra Ready 9/9, no storage-proxy or CSIDriver, retired Volume API returns 404
- rootfs file write -> asynchronous pause -> resume persistence and rootfs snapshot creation
- ctld A/B primary deletion and standby promotion, followed by both DaemonSets and Sandbox0Infra returning Ready


AlexOwl:
Why did you remove it? It was one of the best features of s0 for cargo target and other golden warm cache problems which nor btrfs/zfs/ceph could solve as clean as s0

laotoutou:
> Why did you remove it? It was one of the best features of s0 for cargo target and other golden warm cache problems which nor btrfs/zfs/ceph could solve as clean as s0

I really appreciate your high opinion of s0fs. Could you share the scenarios where you use volumes? 

I removed volumes and s0fs because I realized that volumes are essentially stateful capabilities added to stateless services in the microservices era, whereas the sandbox itself is inherently stateful—here, state refers to the rootfs. 

I'm currently working on decoupling the sandbox rootfs from compute resources, so that the rootfs persists even after the compute resources are destroyed. In other words, the capabilities of volumes will be transferred to the rootfs, which will reduce the complexity of sandbox0 while providing stronger read/write performance for the filesystem. 

Although the current rootfs already achieves this, it's not yet optimal—a major issue is that it doesn't support lazy loading like volumes do. I'm actively working on solving this problem.

